PR #31, opened 2015-01-16
Allow only one layout

jasuarez:
Change Keyboard settings so only one layout can be selected

[endlessm/eos-shell#3364]


rshuler:
As discussed on #3364, we don't want this now, as it will prevent the option of adding more than one layout, which some users will need.  Let's wait for the new upstream solution.

PR #34, opened 2015-02-23
Shell: Removed out the "Suspend and Power Off" block in power panel

sgnn7:
We don't have any user-reachable way to suspend nor hibernate so those
two options have been removed from the power configuration panel.

[endlessm/eos-shell#4625]


cosimoc:
Fine for now.

PR #35, opened 2015-03-12
Add password reminder

jasuarez:
When setting the password, set also the password reminder.

[endlessm/eos-shell#1623]


jasuarez:
Let's make password reminder completely optional in control center.

PR #38, opened 2015-03-31
datetime: Always allow to select a 12-h format, regardless of locale

mariospr:
Always use a locale-specific translation for the AM/PM strings if it's
available, but do not prevent users from using 12-h formats if it's not,
and use a fallback "AM / PM" strings for those situations.

[endlessm/eos-shell#4772]

### DIFF
--- a/panels/datetime/cc-datetime-panel.c
+++ b/panels/datetime/cc-datetime-panel.c
@@ -84,7 +84,7 @@ struct _CcDateTimePanelPrivate
   GSettings *clock_settings;
   GSettings *datetime_settings;
   GDesktopClockFormat clock_format;
-  gboolean ampm_available;
+  gboolean fallback_ampm;
   GtkWidget *am_label;
   GtkWidget *pm_label;
   GtkWidget *am_pm_stack;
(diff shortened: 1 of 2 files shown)


panels/datetime/cc-datetime-panel.c
```diff
@@ -359,8 +369,12 @@ update_time (CcDateTimePanel *self)
   /* Update the time on the listbow row */
   if (use_ampm)
     {
+      gchar *format_string = g_strdup_printf ("%s %s", _("%e %B %Y, %l:%M"),
```
cosimoc: In general these changes look good, but you should try to preserve all the existing translatable strings, instead of changing them.
You can just use the `fallback_ampm` variable and add a new string when it's true, using the old one when it's not.

cosimoc: To be clear, this is true for the other changes in this PR as well.

mariospr: I tested this and I'd swear that the new resultant strings got translated as well, but I see what you mean: just adding the AM/PM string when fallback is true would work, since the translation for %p should be an empty string, so better use that. Will change that.

PR #42, opened 2015-04-23
Use org.freedesktop.login1.Manager.CanSuspend from the power panel

mariospr:
Show "Suspend & Power Off" section only when system can suspend
Instead of disabling this section entirely, ask logind's D-Bus service
'org.freedesktop.login1.Manager.CanSuspend' to determine whether it
should be visible or not, similar to what GDM and the Shell already do.

Also, fix two critical messages while at it.

[endlessm/eos-shell#4655]

### DIFF
--- a/panels/power/cc-power-panel.c
+++ b/panels/power/cc-power-panel.c
@@ -2143,6 +2143,40 @@ add_device_section (CcPowerPanel *self)
   gtk_widget_show_all (box);
 }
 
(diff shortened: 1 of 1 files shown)


mariospr:
> AFAICS the string will be leaked unless you use `g_variant_get (reply, "(&s)", &s)`

You're right. Will change it.

PR #50, opened 2015-08-28
Shell/4019

cosimoc:
endlessm/eos-shell#4019


mcatanzaro:
@cosimoc, this is obsolete as the code lives in gnome-session now, right?

cosimoc:
Indeed, this is not needed anymore.

PR #54, opened 2015-12-12
Issue/6133

GeorgesStavracas:
This pull request solves the main issues with GNOME Control Center running on small screens. The changes:
- Don't make the number of columns in CategoryView class fixed to 6, so icons can adapt to different window sizes.
- Reduce the window's size request to 640, which is the smaller screen size we support.
- Adapt the panels so they don't request too much space.

[endlessm/eos-shell#6133]

### DIFF
--- a/panels/info/cc-info-panel.c
+++ b/panels/info/cc-info-panel.c
@@ -100,6 +100,9 @@ struct _CcInfoPanelPrivate
 {
   GtkBuilder    *builder;
   GtkWidget     *extra_options_dialog;
+  GtkWidget     *scrolled_window;
+  GtkWidget     *detail_vbox;
+  GtkWidget     *hbox1;
 
   GCancellable  *cancellable;
 
@@ -375,6 +378,31 @@ cc_info_panel_finalize (GObject *object)
 }
 
 static void
+cc_info_panel_constructed (GObject *object)
+{
+  CcInfoPanel *self = CC_INFO_PANEL (object);
+  CcShell *shell;
+
+  G_OBJECT_CLASS (cc_info_panel_parent_class)->constructed (object);
+
+  shell = cc_panel_get_shell (CC_PANEL (object));
+
+  if (cc_shell_is_small_screen (shell))
+    {
+      GtkWidget *sw;
+
+      sw = gtk_scrolled_window_new (NULL, NULL);
+      gtk_scrolled_window_set_min_content_height (GTK_SCROLLED_WINDOW (sw), 380);
+
+      gtk_container_remove (GTK_CONTAINER (self->priv->hbox1), self->priv->detail_vbox);
+      gtk_container_add (GTK_CONTAINER (sw), self->priv->detail_vbox);
+      gtk_container_add (GTK_CONTAINER (self->priv->hbox1), sw);
+
+      gtk_widget_show (sw);
+    }
+}
+
+static void
 cc_info_panel_class_init (CcInfoPanelClass *klass)
 {
   GObjectClass *object_class = G_OBJECT_CLASS (klass);
@@ -383,6 +411,7 @@ cc_info_panel_class_init (CcInfoPanelClass *klass)
 
   object_class->dispose = cc_info_panel_dispose;
   object_class->finalize = cc_info_panel_finalize;
+  object_class->constructed = cc_info_panel_constructed;
 }
 
 static char *
@@ -1840,6 +1869,8 @@ cc_info_panel_init (CcInfoPanel *self)
       return;
     }
 
+  self->priv->hbox1 = WID ("hbox1");
+  self->priv->detail_vbox = WID ("detail_vbox");
   self->priv->updater_proxy = eos_updater_proxy_new_for_bus_sync (G_BUS_TYPE_SYSTEM,
                                                                   G_DBUS_PROXY_FLAGS_NONE,
                                                                   "com.endlessm.Updater",
--- a/panels/mouse/gnome-mouse-properties.ui
+++ b/panels/mouse/gnome-mouse-properties.ui
@@ -59,7 +59,7 @@
                         </child>
                       </object>
                       <packing>
-                        <property name="expand">True</property>
+                        <property name="expand">False</property>
                         <property name="fill">True</property>
                         <property name="position">0</property>
                       </packing>
@@ -76,9 +76,8 @@
                           <object class="GtkTable" id="table3">
                             <property name="visible">True</property>
                             <property name="can_focus">False</property>
-                            <property name="margin_right">220</property>
                             <property name="n_rows">2</property>
-                            <property name="n_columns">3</property>
+                            <property name="n_columns">2</property>
                             <property name="column_spacing">40</property>
                             <property name="row_spacing">10</property>
                             <child>
@@ -86,6 +85,7 @@
                                 <property name="visible">True</property>
                                 <property name="can_focus">False</property>
                                 <property name="spacing">5</property>
+                                <property name="halign">start</property>
                                 <child>
                                   <object class="GtkLabel" id="double_click_slow_label">
                                     <property name="visible">True</property>
@@ -109,6 +109,7 @@
                                     <property name="adjustment">adjustment4</property>
                                     <property name="inverted">True</property>
                                     <property name="draw_value">False</property>
+                                    <property name="width_request">100</property>
                                     <child internal-child="accessible">
                                       <object class="AtkObject" id="double_click_scale-atkobject">
                                         <property name="AtkObject::accessible-description" translatable="yes">Double-click timeout</property>
@@ -234,14 +235,14 @@
                         </child>
                       </object>
                       <packing>
-                        <property name="expand">True</property>
+                        <property name="expand">False</property>
                         <property name="fill">True</property>
                         <property name="position">1</property>
                       </packing>
                     </child>
                   </object>
                   <packing>
-                    <property name="expand">True</property>
+                    <property name="expand">False</property>
                     <property name="fill">True</property>
                     <property name="position">0</property>
                   </packing>
@@ -300,8 +301,7 @@
                           <object class="GtkTable" id="table2">
                             <property name="visible">True</property>
                             <property name="can_focus">False</property>
-                            <property name="margin_right">220</property>
-                            <property name="n_columns">3</property>
+                            <property name="n_columns">2</property>
                             <property name="column_spacing">40</property>
                             <property name="row_spacing">5</property>
                             <child>
@@ -351,9 +351,10 @@
                                     <property name="adjustment">adjustment1</property>
                                     <property name="draw_value">False</property>
                                     <property name="value_pos">right</property>
+                                    <property name="width_request">100</property>
                                   </object>
                                   <packing>
-                                    <property name="expand">True</property>
+                                    <property name="expand">False</property>
                                     <property name="fill">True</property>
                                     <property name="position">1</property>
                                   </packing>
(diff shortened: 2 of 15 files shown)


panels/info/cc-info-panel.c
```diff
@@ -1840,6 +1843,24 @@ cc_info_panel_init (CcInfoPanel *self)
       return;
     }
 
+  self->priv->hbox1 = WID ("hbox1");
+  self->priv->detail_vbox = WID ("detail_vbox");
+
+  if (gdk_screen_width () < 1024 || gdk_screen_height () < 600)
```
cosimoc: I wonder if we can't reuse the "small screen" property of CcWindow instead of hardcoding a size here

panels/info/cc-info-panel.c
```diff
@@ -1840,6 +1843,24 @@ cc_info_panel_init (CcInfoPanel *self)
...
+    {
+      GtkWidget *sw;
+
+      sw = gtk_scrolled_window_new (NULL, NULL);
+      gtk_scrolled_window_set_min_content_height (GTK_SCROLLED_WINDOW (sw), 380);
+
+      gtk_container_remove (GTK_CONTAINER (self->priv->hbox1), self->priv->detail_vbox);
```
cosimoc: Won't this destroy the detail_vbox if you don't hold a ref?

panels/info/cc-info-panel.c
```diff
@@ -1840,6 +1843,24 @@ cc_info_panel_init (CcInfoPanel *self)
...
+      gtk_container_add (GTK_CONTAINER (sw), self->priv->detail_vbox);
+      gtk_container_add (GTK_CONTAINER (self->priv->hbox1), sw);
+
+      gtk_widget_show (sw);
+
```
cosimoc: Nit: remove extra line

panels/info/info.ui
```diff
@@ -378,65 +375,65 @@
...
-				    </child>
-				  </object>
-				  <packing>
-				    <property name="left_attach">1</property>
-				    <property name="right_attach">2</property>
-				    <property name="top_attach">1</property>
-				    <property name="bottom_attach">2</property>
-				  </packing>
-				</child>
+            <child>
```
cosimoc: This looks like it's manually edited and lost its original indentation. It would be nice to only see the real diff here.

panels/info/info.ui
```diff
@@ -654,13 +651,13 @@
                               <object class="GtkLabel" id="attribution_label">
                                 <property name="visible">True</property>
                                 <property name="can_focus">False</property>
-				<property name="use_markup">True</property>
-				<property name="xalign">0</property>
-				<property name="margin_left">22</property>
+        <property name="use_markup">True</property>
+        <property name="xalign">0</property>
+        <property name="margin_left">22</property>
```
cosimoc: Ditto

panels/info/info.ui
```diff
@@ -654,13 +651,13 @@
...
                                 <property name="label" translatable="yes">&lt;a href='attribution-link'&gt;Endless Terms of Use&lt;/a&gt;</property>
-				<attributes>
-				  <attribute name="scale" value="0.83"/>
-				</attributes>
+        <attributes>
+          <attribute name="scale" value="0.83"/>
+        </attributes>
```
cosimoc: Ditto

panels/sound/cc-sound-panel.c
```diff
@@ -110,13 +110,30 @@ cc_sound_panel_finalize (GObject *object)
...
         gtk_window_set_default_icon_name ("multimedia-volume-control");
 
         self->control = gvc_mixer_control_new ("GNOME Volume Control Dialog");
         gvc_mixer_control_open (self->control);
+
         self->dialog = gvc_mixer_dialog_new (self->control);
-        gtk_container_add (GTK_CONTAINER (self), GTK_WIDGET (self->dialog));
+        widget = GTK_WIDGET (self->dialog);
+
+        if (gdk_screen_height () < 600) {
```
cosimoc: See other comment about using the small screen property of the CcWindow

panels/sound/cc-sound-panel.c
```diff
@@ -110,13 +110,30 @@ cc_sound_panel_finalize (GObject *object)
...
+                GtkWidget *sw;
+
+                sw = gtk_scrolled_window_new (NULL, NULL);
+                gtk_scrolled_window_set_min_content_height (GTK_SCROLLED_WINDOW (sw), 400);
```
cosimoc: Could use a define

panels/sound/cc-sound-panel.c
```diff
@@ -110,13 +110,30 @@ cc_sound_panel_finalize (GObject *object)
...
+                gtk_widget_show (GTK_WIDGET (sw));
```
cosimoc: No need to cast this

panels/sound/cc-sound-panel.c
```diff
@@ -110,13 +110,30 @@ cc_sound_panel_finalize (GObject *object)
...
+
+                gtk_container_add (GTK_CONTAINER (sw), GTK_WIDGET (self->dialog));
```
cosimoc: You can use `widget` instead of `self->dialog` and avoid a cast

panels/user-accounts/um-user-panel.c
```diff
@@ -1428,6 +1428,22 @@ cc_user_panel_init (CcUserPanel *self)
                 return;
         }
 
+        /* Add scrollbars when screen is too small */
+        if (gdk_screen_width () < 1024 || gdk_screen_height () < 500) {
```
cosimoc: See other comment about using the small screen property

panels/user-accounts/um-user-panel.c
```diff
@@ -1428,6 +1428,22 @@ cc_user_panel_init (CcUserPanel *self)
...
+                GtkWidget *main_user_vbox, *hbox2, *sw;
+
+                sw = gtk_scrolled_window_new (NULL, NULL);
+                gtk_scrolled_window_set_min_content_width (GTK_SCROLLED_WINDOW (sw), 400);
+
+                main_user_vbox = get_widget (d, "main-user-vbox");
+                hbox2 = get_widget (d, "hbox2");
+                gtk_container_remove (GTK_CONTAINER (hbox2), main_user_vbox);
```
cosimoc: See other comment about possibly needing to hold a ref here

shell/cc-shell-category-view.c
```diff
@@ -138,7 +138,6 @@ cc_shell_category_view_constructed (GObject *object)
   gtk_icon_view_set_text_column (GTK_ICON_VIEW (iconview), COL_NAME);
   gtk_icon_view_set_item_width (GTK_ICON_VIEW (iconview), 100);
   cc_shell_item_view_update_cells (CC_SHELL_ITEM_VIEW (iconview));
-  gtk_icon_view_set_columns (GTK_ICON_VIEW (iconview), 6);
```
cosimoc: Does this change have any impact to the UI for normal size screens?

panels/mouse/gnome-mouse-properties.ui
```diff
@@ -76,9 +76,8 @@
                           <object class="GtkTable" id="table3">
                             <property name="visible">True</property>
                             <property name="can_focus">False</property>
-                            <property name="margin_right">220</property>
                             <property name="n_rows">2</property>
-                            <property name="n_columns">3</property>
+                            <property name="n_columns">2</property>
```
cosimoc: Do these changes affect the layout at other resolutions?

panels/info/cc-info-panel.c
GeorgesStavracas: Can you suggest a way to expose that property (technically it's not a property, but whatever)? The only feasible way would be exposing it as another function in CcShell interface, which can break potential third-party plugins.

Also, the 'small-window' property is not enough for the check that is being made here. So either we would have to change the property, or let the extra checks here,

panels/info/cc-info-panel.c
GeorgesStavracas: Since we're retrieving the widget reference from the GtkBuilder instance which created this widget (the WID macro expands into a gtk_builder_get_object), the builder instance owns a reference widget.

panels/info/cc-info-panel.c
cosimoc: I don't think the control center (differently from the settings daemon) has the ability to add third-party plugins. Either way, we don't ship them in Endless so it wouldn't be a concern.
It should be fine to add a method to the CcWindow API to retrieve the "small window" value.
Also yeah feel free to expand what the "small-window" property checks for, if what's there is not enough (but I'm curious to understand why it's not enough in that case).

panels/info/cc-info-panel.c
cosimoc: Got it, makes sense.

panels/info/cc-info-panel.c
GeorgesStavracas: You're right, I was wrong here. Fixing it.

panels/mouse/gnome-mouse-properties.ui
GeorgesStavracas: I don't think so. Tested with multiple resolutions and it seems to work.

shell/cc-shell-category-view.c
GeorgesStavracas: I don't think so. Tested with multiple resolutions in a VM, and it seems to work as usual.

shell/cc-shell.c
```diff
@@ -171,3 +171,26 @@ cc_shell_embed_widget_in_header (CcShell *shell, GtkWidget *widget)
...
+gboolean
+cc_shell_is_small_screen (CcShell *shell)
+{
+  CcShellInterface *iface;
+
+  g_return_if_fail (CC_IS_SHELL (shell));
+
+  iface = CC_SHELL_GET_IFACE (shell);
+
+  if (!iface->embed_widget_in_header)
```
cosimoc: Should be `iface->is_small_screen`

shell/cc-window.c
```diff
@@ -1168,12 +1170,21 @@ cc_window_finalize (GObject *object)
   G_OBJECT_CLASS (cc_window_parent_class)->finalize (object);
 }
 
+static gboolean
+_shell_is_small_screen (CcShell *shell)
+{
+  update_small_screen_settings (CC_WINDOW (shell));
```
cosimoc: Do you really need to call this every time?

panels/sound/cc-sound-panel.c
```diff
@@ -116,7 +142,9 @@ cc_sound_panel_init (CcSoundPanel *self)
 
         self->control = gvc_mixer_control_new ("GNOME Volume Control Dialog");
         gvc_mixer_control_open (self->control);
+
```
cosimoc: No need to add these empty lines

panels/sound/cc-sound-panel.c
```diff
@@ -70,6 +70,31 @@ cc_sound_panel_set_property (GObject      *object,
...
+        CcShell *shell;
+
+        shell = cc_panel_get_shell (CC_PANEL (self));
+
+        G_OBJECT_CLASS (cc_sound_panel_parent_class)->constructed (object);
+
+        if (cc_shell_is_small_screen (shell)) {
+                GtkWidget *sw;
+
+                g_object_ref (self->dialog);
```
cosimoc: Don't you need to unref this later?

shell/cc-window.c
```diff
@@ -117,6 +117,8 @@ static gboolean cc_window_set_active_panel_from_id (CcShell      *shell,
 
 static gint get_monitor_height (CcWindow *self);
 
+static void update_small_screen_settings (CcWindow *self);
```
cosimoc: Not sure you still need this - no need for another round of review though for it.

PR #58, opened 2016-01-07
Issue/6133

GeorgesStavracas:
Fix the following panels:
 - Account
 - Background
 - Region & language

[endlessm/eos-shell#6133]

### DIFF
--- a/panels/region/cc-input-chooser.c
+++ b/panels/region/cc-input-chooser.c
@@ -1211,3 +1211,21 @@ cc_input_chooser_get_selected (GtkWidget  *chooser,
 
   return TRUE;
 }
+
+void
+cc_input_chooser_set_is_small_screen (GtkWidget *chooser,
+                                      gboolean   is_small_screen)
+{
+        CcInputChooserPrivate *priv = GET_PRIVATE (chooser);
+
+        if (is_small_screen) {
+                guint width, height;
+
+                gtk_window_get_size (gtk_window_get_transient_for (GTK_WINDOW (chooser)), &width, &height);
+                gtk_widget_set_size_request (GTK_WIDGET (chooser), width * MAIN_WINDOW_WIDTH_RATIO, height);
+
+                gtk_scrolled_window_set_policy (GTK_SCROLLED_WINDOW (priv->scrolledwindow),
+                                                GTK_POLICY_AUTOMATIC,
+                                                GTK_POLICY_AUTOMATIC);
+        }
+}
--- a/panels/user-accounts/um-account-dialog.c
+++ b/panels/user-accounts/um-account-dialog.c
@@ -66,6 +66,7 @@ static void   um_account_dialog_response  (GtkDialog *dialog,
 struct _UmAccountDialog {
         GtkDialog parent;
         GtkWidget *container_widget;
+        GtkWidget *scrolledwindow;
         GSimpleAsyncResult *async;
         GCancellable *cancellable;
         GPermission *permission;
@@ -1486,6 +1487,8 @@ um_account_dialog_init (UmAccountDialog *self)
         gtk_container_add (GTK_CONTAINER (content), widget);
         self->container_widget = widget;
 
+        self->scrolledwindow = (GtkWidget *) gtk_builder_get_object (builder, "account-dialog-scrolledwindow");
+
         local_init (self, builder);
         enterprise_init (self, builder);
         join_init (self, builder);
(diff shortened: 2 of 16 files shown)


panels/region/cc-input-chooser.c
```diff
@@ -1211,3 +1211,30 @@ cc_input_chooser_get_selected (GtkWidget  *chooser,
...
+cc_input_chooser_set_is_small_screen (GtkWidget *chooser,
+                                      gboolean   is_small_screen)
+{
+        CcInputChooserPrivate *priv = GET_PRIVATE (chooser);
+
+        if (is_small_screen) {
+                guint width, height;
+
+                gtk_window_get_size (gtk_window_get_transient_for (GTK_WINDOW (chooser)), &width, &height);
+                gtk_widget_set_size_request (GTK_WIDGET (chooser), width * MAIN_WINDOW_WIDTH_RATIO, height);
```
cosimoc: Is this really needed? Can you not just let the dialog size the way it normally would?

panels/user-accounts/um-account-dialog.c
```diff
@@ -1456,6 +1457,8 @@ um_account_dialog_init (UmAccountDialog *self)
         gtk_window_set_title (GTK_WINDOW (dialog), _("Add User"));
         gtk_window_set_icon_name (GTK_WINDOW (dialog), "system-users");
 
+  g_message ("content: %p", content);
```
cosimoc: Remove this

panels/user-accounts/um-account-dialog.c
```diff
@@ -1678,3 +1683,21 @@ um_account_dialog_finish (UmAccountDialog     *self,
...
+                                       gboolean         is_small_screen)
+{
+        if (is_small_screen) {
+                GtkWindow *transient;
+                gint width, height;
+
+                transient = gtk_window_get_transient_for (GTK_WINDOW (self));
+
+                gtk_window_get_size (transient, &width, &height);
+                gtk_widget_set_size_request (GTK_WIDGET (self), width, height);
```
cosimoc: Again, is this really needed?

panels/user-accounts/um-account-dialog.c
```diff
@@ -1678,3 +1683,21 @@ um_account_dialog_finish (UmAccountDialog     *self,
         g_clear_object (&self->async);
         return user;
 }
+
+void
+um_account_dialog_set_is_small_screen (UmAccountDialog *self,
+                                       gboolean         is_small_screen)
+{
+        if (is_small_screen) {
```
cosimoc: Do we need to do anything in the `!is_small_screen` case?

panels/region/cc-input-chooser.c
GeorgesStavracas: No, it's needed. Because of the scrolled window, the dialog would shrink to a minimum size, which is quite small. So, something has to have a size request, and I set the dialog's one (although it could be the scrolled window's size request too).

panels/user-accounts/um-account-dialog.c
GeorgesStavracas: Ops!

panels/user-accounts/um-account-dialog.c
GeorgesStavracas: Ditto.

panels/user-accounts/um-account-dialog.c
GeorgesStavracas: Don't think so. We usually call this function only once, since the dialogs are usually recreated when needed.

panels/region/cc-input-chooser.c
cosimoc: Would something like `gtk_scrolled_window_set_min_content_width` help?

panels/user-accounts/um-account-dialog.c
cosimoc: The reason I was asking is because the other similar functions you add for the dialogs take both cases into account.
I guess if we never expect this to change during runtime, those other cases can be made simpler too?

panels/region/cc-input-chooser.c
GeorgesStavracas: I personally prefer to stay with the transient's geometry. Using the scrolled-window's min_content_width & min_content_height would mean that we'll have to use arbitrary hardcoded sizes. Also, setting the scrolledwindow's size won't take into account the buttons and other misc margins and paddings.

By setting the dialog's size-request, we avoid all the previous issues and guarantee that everything is visible from a small screen.

panels/user-accounts/um-account-dialog.c
GeorgesStavracas: Yeah, we're not expecting runtime changes. Just removed the code from the other cases.

PR #65, opened 2016-05-16
T3597

mcatanzaro:


mcatanzaro:
https://phabricator.endlessm.com/T3597

panels/user-accounts/um-utils.c
```diff
@@ -460,24 +460,8 @@ down_arrow (GtkStyleContext *context,
...
-                g_assert_cmpint (length, >=, 0);
-        }
-
-        return length;
-#endif
-}
-
-#define MAXNAMELEN  get_login_name_max ()
+/* This is best hardcoded, see GNOME #766401. */
+#define MAXNAMELEN  32
```
cosimoc: I see that in the upstream patch you took a different approach and still consult `sysconf`. Also Matthias had some comments on the patch. I would prefer if we followed upstream here, and if this is not really crucial to the patchset, I'd rather we branched it off in another issue (or use an upstream patch)

mcatanzaro: I think it's OK to deal with this in a separate issue, but it really needs to be fixed in the same release, else it will be very hard to figure out why user account creation is failing. It seems easiest to track it in the same issue.

I don't know how to respond to Matthias's review, so I filed a bug against shadow-utils. Let's see if they respond and what they think.

PR #80, opened 2016-12-28
display-panel: don't show overscan toggle when unsupported

GeorgesStavracas:
When the overscan feature is not supported by the output, we
still show an useless checkbox that does nothing and has a high
potential to confuse users.

Fix that by only showing the checkbox when the overscan feature
is supported.

https://phabricator.endlessm.com/T13812


rshuler:
Does the commit message refer to the correct task?

PR #81, opened 2017-02-27
T15794

cosimoc:
https://phabricator.endlessm.com/T15794

### DIFF
--- a/panels/info/cc-info-panel.c
+++ b/panels/info/cc-info-panel.c
@@ -122,6 +122,8 @@ struct _CcInfoPanelPrivate
 };
 
 static void get_primary_disc_info_start (CcInfoPanel *self);
+static void sync_state_from_updater     (CcInfoPanel *self,
+                                         gboolean is_initial_state);
 
 typedef struct
 {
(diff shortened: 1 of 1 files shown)


panels/info/cc-info-panel.c
```diff
@@ -1469,7 +1471,12 @@ updates_apply_completed (GObject *object,
   if (error != NULL)
     {
       if (!g_error_matches (error, G_IO_ERROR, G_IO_ERROR_CANCELLED))
-        g_warning ("Failed to call Apply() on EOS updater: %s", error->message);
+        {
+          CcInfoPanel *self = user_data;
+          g_warning ("Failed to call Apply() on EOS updater: %s", error->message);
+          sync_state_from_updater (self, FALSE);
```
pwithnall: I’m not sure how this will indicate a local error in the UI — it seems to just query the updater state for the remote error. What am I missing?

pwithnall:
Everything else looks good.

panels/info/cc-info-panel.c
cosimoc: Yeah, the commit message is a bit misleading -- that patch will effectively make it so when the daemon is already in an error state, and the call fails, such error state is propagated to the UI, undo-ing the initial state setup we do here: https://github.com/endlessm/gnome-control-center/blob/master/panels/info/cc-info-panel.c#L1581

I could write another follow-up patch that also considers the local error, regardless of what state the daemon is in, if you think it's worth it.

pwithnall: Since a local error is pretty unlikely (assuming the system is properly set up, which we can assume), I don’t think it’s necessary to write the follow-up patch now. It’s probably a good idea to file a follow-up task about it though.

PR #82, opened 2017-03-09
Allow defining branding-specific logos for the "info" panel.

mariospr:
If a configuration file in /etc/eos-branding/gnome-control-center/info.conf
is found, it defines the path to a branding-specific logo and such a file
exists in disk, then gnome-control-center will use that one to override
Endless's logo in the "info" panel.

https://phabricator.endlessm.com/T15836

### DIFF
--- a/panels/info/cc-info-panel.c
+++ b/panels/info/cc-info-panel.c
@@ -56,6 +56,21 @@
 
 #define MEDIA_HANDLING_SCHEMA "org.gnome.desktop.media-handling"
 
+/*
+ * CONFIG_FILEPATH points to /etc/gnome-control-center/branding.conf, which
+ * is a file containing groups of key-value pairs under a group called "Info", matching
+ * the name of this panel.
+ *
+ * Fields:
+ *   - logo (required): absolute path to the file with a logo for the brand.
+ *
+ * This is how this file would look on a branded image:.
+ *
+ *   [Info]
+ *   logo=/path/to/the/image/with/the/logo.png
+ */
+#define CONFIG_FILEPATH CONFIGDIR "/branding.conf"
+
 #define WID(w) (GtkWidget *) gtk_builder_get_object (self->priv->builder, w)
 
 CC_PANEL_REGISTER (CcInfoPanel, cc_info_panel)
@@ -103,6 +118,8 @@ struct _CcInfoPanelPrivate
   GtkWidget     *scrolled_window;
   GtkWidget     *detail_vbox;
   GtkWidget     *hbox1;
+  GtkWidget     *logo_image;
+  gchar         *branding_logo_path;
 
   GCancellable  *cancellable;
 
@@ -276,6 +293,8 @@ static void
 cc_info_panel_finalize (GObject *object)
 {
   CcInfoPanelPrivate *priv = CC_INFO_PANEL (object)->priv;
+
+  g_free (priv->branding_logo_path);
 
   if (priv->cancellable)
     {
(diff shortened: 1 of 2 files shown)


panels/info/cc-info-panel.c
```diff
@@ -1821,6 +1906,11 @@ cc_info_panel_init (CcInfoPanel *self)
 
   self->priv->hbox1 = WID ("hbox1");
   self->priv->detail_vbox = WID ("detail_vbox");
+  self->priv->logo_image = WID ("system_image");
+  self->priv->branding_logo_path = NULL;
```
cosimoc: No need to initialize this explicitly.

panels/info/cc-info-panel.c
```diff
@@ -1798,6 +1817,72 @@ info_panel_setup_overview (CcInfoPanel  *self)
...
+  g_strfreev (groups);
+  g_key_file_free (keyfile);
+
+  return ret;
+}
+
+static void
+update_branding_specific_info (CcInfoPanel *self)
+{
+  if (!read_config_file (self)) {
```
cosimoc: I would change `read_config_file()` to not return any value, and move the debug message inside the next if. Basically it does not make sense to check for both the function returning false and the logo path being NULL.

panels/info/cc-info-panel.c
```diff
@@ -1798,6 +1817,72 @@ info_panel_setup_overview (CcInfoPanel  *self)
...
+{
+  GKeyFile *keyfile = NULL;
+  GError *error = NULL;
+  gchar **groups = NULL;
+  gboolean ret = FALSE;
+  gsize n_groups;
+  int i;
+
+  if (!g_file_test (CONFIG_FILEPATH, G_FILE_TEST_EXISTS))
+    return FALSE;
```
cosimoc: Fold this condition when looking at the error code returned by `g_key_file_load_from_file()`

panels/info/cc-info-panel.c
```diff
@@ -1798,6 +1817,72 @@ info_panel_setup_overview (CcInfoPanel  *self)
...
+
+  keyfile = g_key_file_new ();
+  if (!g_key_file_load_from_file (keyfile, CONFIG_FILEPATH, G_KEY_FILE_NONE, &error))
+    {
+      g_warning ("Could not read file %s: %s\n", CONFIG_FILEPATH, error->message);
```
cosimoc: `g_warning()` already puts a newline (multiple occurrences)

panels/info/cc-info-panel.c
```diff
@@ -1798,6 +1817,72 @@ info_panel_setup_overview (CcInfoPanel  *self)
...
+    }
+
+  groups = g_key_file_get_groups (keyfile, &n_groups);
+  for (i = 0; i < n_groups; i++)
+    {
+      size_t len;
+
+      /* We only support "Info" group here */
+      if (g_strcmp0 (groups[i], "Info") != 0)
+        continue;
```
cosimoc: Just change the code to directly look at the `Info` group.

panels/info/cc-info-panel.c
```diff
@@ -1798,6 +1817,72 @@ info_panel_setup_overview (CcInfoPanel  *self)
...
+    {
+      g_warning ("Could not read file %s: %s\n", CONFIG_FILEPATH, error->message);
+      g_error_free (error);
+      goto out;
+    }
+
+  groups = g_key_file_get_groups (keyfile, &n_groups);
+  for (i = 0; i < n_groups; i++)
+    {
+      size_t len;
```
cosimoc: Unused variable

panels/info/cc-info-panel.c
mariospr: Ok

panels/info/cc-info-panel.c
mariospr: Opps!

panels/info/cc-info-panel.c
mariospr: I'm going to remove this loop entirely, so this will go away

panels/info/cc-info-panel.c
mariospr: Yep

panels/info/cc-info-panel.c
mariospr: Good point, I'll change that

panels/info/cc-info-panel.c
mariospr: ok

PR #87, opened 2017-06-09
T17270

joaquimrocha:
https://phabricator.endlessm.com/T17270

### DIFF
--- a/panels/user-accounts/data/account-dialog.ui
+++ b/panels/user-accounts/data/account-dialog.ui
@@ -278,7 +278,7 @@
                     </child>
                     <child>
                       <object class="GtkRadioButton" id="local_no_password_radio">
-                        <property name="label" translatable="yes">Allow user to log in _without a password</property>
+                        <property name="label" translatable="yes">Do not re_quire a password when logging in</property>
                         <property name="visible">True</property>
                         <property name="can_focus">True</property>
                         <property name="receives_default">False</property>
--- a/panels/user-accounts/data/account-dialog.ui
+++ b/panels/user-accounts/data/account-dialog.ui
@@ -277,6 +277,22 @@
                       </packing>
                     </child>
                     <child>
+                      <object class="GtkRadioButton" id="local_no_password_radio">
+                        <property name="label" translatable="yes">Allow user to log in _without a password</property>
+                        <property name="visible">True</property>
+                        <property name="can_focus">True</property>
+                        <property name="receives_default">False</property>
+                        <property name="active">True</property>
+                        <property name="draw_indicator">True</property>
+                        <property name="group">local_password_login_radio</property>
+                        <property name="use_underline">True</property>
+                        <property name="margin_start">12</property>
+                      </object>
+                      <packing>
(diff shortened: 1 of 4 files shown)


panels/user-accounts/data/password-dialog.ui
```diff
@@ -249,6 +249,20 @@
                       </packing>
                     </child>
                     <child>
+                      <object class="GtkRadioButton" id="action-no-password-radio">
+                        <property name="label" translatable="yes">Allow user to login without a password</property>
```
mcatanzaro: "log in" should be two words here since it's used as a verb, unlike "on next login" where it is used as a noun (one word).

panels/user-accounts/um-password-dialog.c
```diff
@@ -219,10 +220,14 @@ accept_password_dialog (GtkButton        *button,
                         break;
 
                 case ACT_USER_PASSWORD_MODE_SET_AT_LOGIN:
-                        act_user_set_password_mode (um->user,  um->password_mode);
                         act_user_set_automatic_login (um->user, FALSE);
+                case ACT_USER_PASSWORD_MODE_NONE:
```
mcatanzaro: I would use a break statement here, even if it's not strictly necessary and adds a bit more code, for readability and robustness to future modification, including possible upstream modification.

panels/user-accounts/data/account-dialog.ui
```diff
@@ -277,6 +277,22 @@
                       </packing>
                     </child>
                     <child>
+                      <object class="GtkRadioButton" id="local_no_password_radio">
+                        <property name="label" translatable="yes">Allow user to login _without a password</property>
```
mcatanzaro: log in (two words)

joaquimrocha: Thought so too but there's already:
_Allow user to set a password when they next login_

mcatanzaro: Hmm you're right... that label is wrong too. :)

panels/user-accounts/um-password-dialog.c
```diff
@@ -223,6 +224,14 @@ accept_password_dialog (GtkButton        *button,
                         act_user_set_automatic_login (um->user, FALSE);
                         break;
 
+                case ACT_USER_PASSWORD_MODE_NONE:
+                        /* When setting the user password mode to none, when it
+                         * was previously set-at-login, we need to reset the password,
+                         * otherwise the mode will be set back to set-at-login */
+                        act_user_set_password (um->user, "", "");
+                        act_user_set_password_mode (um->user,  um->password_mode);
```
mcatanzaro: Nit: I would probably write ACT_USER_PASSWORD_MODE_NONE here, rather than um->password_mode, for clarity. Doesn't matter.

joaquimrocha: Well, that makes the code more orthogonal without a much improved readability IMO: if you have to change the case later you also have to change that call. No need for repetition.

PR #93, opened 2017-08-17
shell: send a metric event when a panel is opened

cosimoc:
https://phabricator.endlessm.com/T18443


shell/cc-window.c
```diff
@@ -117,6 +118,17 @@ static gboolean cc_window_set_active_panel_from_id (CcShell      *shell,
 
 static gint get_monitor_height (CcWindow *self);
 
+#define PANEL_OPENED_EVENT_ID "3c5d59d2-6c3f-474b-95f4-ac6fcc192655"
```
wjt: Please update https://phabricator.endlessm.com/w/software/metrics_events/

wjt: (I am doing this now.)

PR #100, opened 2017-10-13
debian: Use autogen.sh instead of dh_auto_configure

pwithnall:
dh_auto_configure defaults to running configure, which means we can’t
easily build the package from git.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -12,6 +12,7 @@ override_dh_autoreconf:
 	dh_autoreconf --as-needed
 
 override_dh_auto_configure:
+	NOCONFIGURE=1 ./autogen.sh
 	dh_auto_configure -- \
 		--libexecdir=\$${prefix}/lib/gnome-control-center \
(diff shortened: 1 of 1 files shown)


pwithnall:
> Approved, but would it be worth submitting this as a proposal to Debian?

I’m not sure it’s worth it for such a small patch, given that I’m not a DD and have never interacted with Debian people. If someone within Endless who *is* more of a Debianista than me wants to submit it, that would be great.

rshuler:
Gnome-control-center is failing to build on OBS, and I suspect this may be responsible.

pwithnall:
(It wasn’t responsible: https://phabricator.endlessm.com/T19670)

rshuler:
Actually, I think this was responsible for the build failure on OBS.  T19670 was a separate issue that showed up on jenkins, and only on master.  IIRC, this was failing on 3.3 on OBS, and was fixed by reverting this commit.

pwithnall:
Correct. Sorry, I missed ‘OBS’ when reading your comment.

PR #101, opened 2017-10-13
Fix calculation of total disk size in info panel

pwithnall:

This fixes the calculation of the total disk size in the info panel to not be 5× what it should be. With this set of patches applied, my test VM correctly shows a disk size of 63.6GB. `df -h` shows 60GB, but the difference can be exactly explained away as the difference between base-2 and base-10 gigabytes. (Sigh.)

These are several patches in this branch:
 - “info: Fix total disc size for btrfs subvolumes” is the upstream fix needed for the disk size calculation
 - Various other “info: *” patches fix memory leaks or crashes in the info plugin, and they looked like a good idea to backport
 - “info: Ignore more virtual file systems” is not technically needed for the total disk size fix, but does speed up the panel load (since we’re now querying the size of ~5 file systems rather than 25)


panels/info/cc-info-panel.c
```diff
@@ -585,6 +582,7 @@ get_primary_disc_info (CcInfoPanel *self)
   g_list_free (points);
   g_hash_table_destroy (hash);
 
+  self->priv->cancellable = g_cancellable_new ();
```
mariospr: Why not doing this in `cc_info_panel_init ()`?

pwithnall: Avoids allocating it until the disk info is actually queried. Currently, that’s inevitable, but I guess it could change in future. I don’t want our copy of this patch to diverge from upstream, though, since it’s a simple cherry pick from upstream.

mariospr: Fair enough. I did realize it was from upstream, just curious on the rationale.

PR #102, opened 2017-10-17
Revert "debian: Use autogen.sh instead of dh_auto_configure"

pwithnall:
This reverts commit 50fd5284b998e6642828af28dfbf7c4b4bed0942.

The upstream tarball does not contain an autogen.sh script, so the
package build fails when building from a tarball rather than from git.
Since we build from tarballs, this is embarrassing.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -12,7 +12,6 @@ override_dh_autoreconf:
 	dh_autoreconf --as-needed
 
 override_dh_auto_configure:
-	NOCONFIGURE=1 ./autogen.sh
 	dh_auto_configure -- \
 		--libexecdir=\$${prefix}/lib/gnome-control-center \
 		--with-gnome-session-libexecdir=\$${prefix}/lib/gnome-session \


mariospr:
I have to say I was surprised when I reviewed because I didn't see the NOCONFIGURE thing in any package so far in endless, but because it's a common thing to do anyway I thought it was fine and didn't double check that.

Long way of saying that... can we please share the embarrassment feeling? :-)

pwithnall:
> I have to say I was surprised when I reviewed because I didn't see the NOCONFIGURE thing in any package so far in endless, but because it's a common thing to do anyway I thought it was fine and didn't double check that.

It is a common thing; the problem here is not the `NOCONFIGURE` variable, it’s the fact that the `autogen.sh` script is not included in the upstream tarball of gnome-control-center (which it should be — I’ve filed https://bugzilla.gnome.org/show_bug.cgi?id=789091 about including it in future).

PR #103, opened 2017-11-22
info: use UDisks to retrieve disk size

GeorgesStavracas:
The current code relies on GLib API and uses the
available mounts to calculate the available partition
size. This is because this code assumes that more
than one OS can be installed in the same drive, and
wouldn't make sense to show the whole disk size in
this situation.

However, Endless OS is not meant to be installed with
other OSes on different partitions, and for our partners
it might be bad to show a disk space smaller than what
they put on their hardware.

Fix that by using the UDisks API to get the real size
of the full disks.

https://phabricator.endlessm.com/T19907

### DIFF
--- a/panels/info/cc-info-panel.c
+++ b/panels/info/cc-info-panel.c
@@ -35,6 +35,7 @@
 #include <glibtop/mountlist.h>
 #include <glibtop/mem.h>
 #include <glibtop/sysinfo.h>
+#include <udisks/udisks.h>
 
 #ifdef GDK_WINDOWING_WAYLAND
 #include <gdk/gdkwayland.h>
@@ -110,10 +111,6 @@ struct _CcInfoPanelPrivate
 
   GCancellable  *cancellable;
 
-  /* Free space */
-  GList         *primary_mounts;
-  guint64        total_bytes;
-
   /* Media */
   GSettings     *media_settings;
   GtkWidget     *other_application_combo;
@@ -123,10 +120,10 @@ struct _CcInfoPanelPrivate
   EosUpdater *updater_proxy;
   GDBusProxy *session_proxy;
   GCancellable *updater_cancellable;
+  UDisksClient *client;
   gboolean updater_activated;
 };
 
-static void get_primary_disc_info_start (CcInfoPanel *self);
 static void sync_state_from_updater     (CcInfoPanel *self,
                                          gboolean is_initial_state);
 static void sync_initial_state_from_updater (CcInfoPanel *self);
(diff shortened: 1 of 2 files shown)


panels/info/cc-info-panel.c
```diff
@@ -1939,6 +1876,15 @@ cc_info_panel_init (CcInfoPanel *self)
       g_error_free (error);
     }
 
+  self->priv->client = udisks_client_new_sync (NULL, &error);
+
+  if (error != NULL)
+    {
+      g_critical ("Unable to get UDisks client: %s. Disk information will not be available.",
```
pwithnall: I’d use `g_warning()` here just in case criticals are set to be fatal. We probably don’t want this to be fatal.

panels/info/cc-info-panel.c
```diff
@@ -484,117 +479,59 @@ get_os_description (void)
...
+      gtk_label_set_text (GTK_LABEL (label), size);
+    }
+  else
+    {
+      gtk_label_set_text (GTK_LABEL (label), "-");
+    }
+
+  g_list_free_full (objects, g_object_unref);
+  g_clear_pointer (&added_drives, g_hash_table_destroy);
+  g_clear_object (&manager);
```
pwithnall: Nitpick: You can use `g_autoptr()` on the declarations to avoid having to do this memory management manually. (g-c-c depends on GLib 2.44, which is when `g_autoptr()` was introduced). What’s here is correct though.

Note: When I say ‘nitpick’ in a review, that means it’s something which is not a blocker for the PR being accepted. Fix it if you want, or justify not fixing it or deferring a fix — either is fine.

panels/info/cc-info-panel.c
```diff
@@ -484,117 +479,59 @@ get_os_description (void)
...
-
-static void
-get_primary_disc_info (CcInfoPanel *self)
-{
-  GList        *points;
-  GList        *p;
-  GHashTable *hash;
+  manager = udisks_client_get_object_manager (self->priv->client);
+  objects = g_dbus_object_manager_get_objects (manager);
+  added_drives = g_hash_table_new (g_str_hash, g_str_equal);
```
pwithnall: Nitpick: I’d always use `g_hash_table_new_full()` instead, and pass `NULL` for the `GDestroyNotify` arguments to make it more explicit that the stored keys and values do not need to be freed.

panels/info/cc-info-panel.c
```diff
@@ -1939,6 +1876,15 @@ cc_info_panel_init (CcInfoPanel *self)
       g_error_free (error);
     }
 
+  self->priv->client = udisks_client_new_sync (NULL, &error);
```
pwithnall: Nitpick: Given that all the initialisation code here is sync, it’s alright to make this call sync as well — but what’s the startup time of the info panel? Is there a noticeable delay when it’s opened? If so, we might want to do some follow-up work to make this async.

GeorgesStavracas: With my Endless One, I couldn't notice any perceivable increase in initialization time. I'd say, it's safe.

PR #105, opened 2017-12-07
user panel, password: add fallback kbd layout if necessary

erikos:
user panel, password: add fallback kbd layout if necessary

If the current keyboard layout is not a latin one we will add the fallback kbd layout 'en'. This calls out to gnome-shell where the logic to detect the keyboard layout is kept. The layout will be available in the popup switcher and the taskbar keyboard layout switcher.

https://phabricator.endlessm.com/T18598

### DIFF
--- a/panels/user-accounts/um-password-dialog.c
+++ b/panels/user-accounts/um-password-dialog.c
@@ -25,6 +25,7 @@
 #include <sys/types.h>
 #include <sys/wait.h>
 
+#include <gio/gio.h>
 #include <glib.h>
 #include <glib/gi18n.h>
 #include <gtk/gtk.h>
@@ -54,6 +55,8 @@ struct _UmPasswordDialog {
 
         ActUser *user;
         ActUserPasswordMode password_mode;
+
+        GDBusProxy *input_source_manager;
 
         GtkWidget *old_password_label;
         GtkWidget *old_password_entry;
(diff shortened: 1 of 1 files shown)


panels/user-accounts/um-password-dialog.c
```diff
@@ -566,6 +625,18 @@ um_password_dialog_new (void)
...
+
+        g_dbus_proxy_new_for_bus (G_BUS_TYPE_SESSION,
+                                  G_DBUS_PROXY_FLAGS_NONE,
+                                  NULL,
+                                  "org.gnome.Shell.InputSourceManager",
+                                  "/org/gnome/Shell/InputSourceManager",
+                                  "org.gnome.Shell.InputSourceManager",
+                                  um->cancellable,
+                                  (GAsyncReadyCallback) input_source_manager_proxy_ready,
+                                  um);
```
cosimoc: It's a bit dangerous that this is created asynchronously, because a caller can show the password dialog before it gets created, which would fail the call to set the password mode. I'd probably make this synchronous.

panels/user-accounts/um-password-dialog.c
```diff
@@ -566,6 +625,18 @@ um_password_dialog_new (void)
 
         g_object_unref (builder);
 
+        um->cancellable = g_cancellable_new ();
```
cosimoc: It may not matter if you do the sync creation, but otherwise this should be cancelled on finalize (and unreffed, otherwise it gets leaked)

panels/user-accounts/um-password-dialog.c
```diff
@@ -115,13 +141,25 @@ cancel_password_dialog (GtkButton        *button,
                         UmPasswordDialog *um)
 {
         finish_password_change (um);
+        shell_enable_password_mode (um, FALSE);
```
cosimoc: No need to call this here, since `finish_password_change()` already turns it off

panels/user-accounts/um-password-dialog.c
```diff
@@ -55,6 +56,9 @@ struct _UmPasswordDialog {
         ActUser *user;
         ActUserPasswordMode password_mode;
 
+        GDBusProxy *input_source_manager;
```
cosimoc: This is leaked

erikos: Good catch, done.

PR #106, opened 2018-01-02
info: Check if drive is both removable and media-removable

GeorgesStavracas:
Checking if a given drive is removable is not enough, because eMMC
drives are legitimately removable, and also because the DBus 'Removable'
property is essentially a guess and should not be used alone.

This commit, then, only considers drives that are ejectable, removable
and media-removable simultaneously when deciding which drives to skip.

https://phabricator.endlessm.com/T20504


dsd:
retest this please

PR #107, opened 2018-01-03
panels: Add content sharing options to sharing dialog

smspillaz:
https://phabricator.endlessm.com/T20381


panels/sharing/cc-sharing-panel.c
```diff
@@ -170,6 +174,13 @@ cc_sharing_panel_dispose (GObject *object)
...
     }
 
+  if (priv->content_sharing_dialog)
+    {
+      gtk_widget_destroy (priv->content_sharing_dialog);
+      priv->content_sharing_dialog = NULL;
+    }
+
+
   g_clear_object (&priv->sharing_proxy);
```
cosimoc: Should also clear the companion app helper proxy

panels/sharing/cc-sharing-panel.c
```diff
@@ -646,6 +657,132 @@ cc_sharing_panel_setup_media_sharing_dialog (CcSharingPanel *self)
...
+  g_variant_unref (discoverable_prop);
+
+  w = GTK_WIDGET (gtk_builder_get_object (self->priv->builder, "content-sharing-switch"));
+  gtk_switch_set_active (GTK_SWITCH (w), content_sharing_enabled);
+}
+
+static gboolean
+cc_sharing_panel_check_content_sharing_available (void)
+{
+  FlatpakInstallation *installation = flatpak_installation_new_system (NULL, NULL);
```
cosimoc: This can fail too

panels/sharing/cc-sharing-panel.c
```diff
@@ -646,6 +657,132 @@ cc_sharing_panel_setup_media_sharing_dialog (CcSharingPanel *self)
...
+      ret = FALSE;
+      g_clear_error (&error);
+    }
+
+  g_object_unref (ref);
+  g_object_unref (installation);
+
+  return ret;
+}
+
```
cosimoc: Remove extra line

panels/sharing/cc-sharing-panel.c
```diff
@@ -646,6 +657,132 @@ cc_sharing_panel_setup_media_sharing_dialog (CcSharingPanel *self)
...
+                                                                                "com.endlessm.CompanionAppServiceAvahiHelper",
+                                                                                "/com/endlessm/CompanionAppServiceAvahiHelper",
+                                                                                "com.endlessm.CompanionApp.AvahiHelper",
+                                                                                NULL, &error);
+  if (error != NULL)
+    {
+      g_critical ("Unable to create a DBus proxy for the companion app avahi helper: %s", error->message);
+      g_error_free (error);
+
+      content_sharing_enabled = FALSE;
```
cosimoc: In this case, you probably also need to make the switch insensitive or something like that -- otherwise we'll crash when trying to access the proxy pointer in the switch callback.
Or even better, fold in this check in the function that determines the visibility of the row; if we fail to create the proxy, it's probably best to hide the whole thing altogether.

panels/sharing/cc-sharing-panel.c
```diff
@@ -646,6 +660,154 @@ cc_sharing_panel_setup_media_sharing_dialog (CcSharingPanel *self)
...
+                                          GParamSpec *pspec,
+                                          CcSharingPanel *self)
+{
+  GError *error = NULL;
+
+  if (gtk_switch_get_active (widget))
+    {
+      g_dbus_proxy_call_sync (self->priv->companion_app_avahi_helper_proxy,
+                              "EnterDiscoverableMode",
+                              g_variant_new ("(i)", -1, "Endless Computer"),
```
cosimoc: If anything, this should be `g_variant_new ("(s)", "Endless Computer")`, but I think we should just get rid of the name parameter, as mentioned in the other PR.

cosimoc: Note that if we do that, it will also be much easier to either make `Discoverable` a writable property, or to provide a `SetDiscoverable()` method that takes a boolean, instead of two separate ones.

panels/sharing/cc-sharing-panel.c
cosimoc: Still needs to be removed.

panels/sharing/cc-sharing-panel.c
```diff
@@ -646,6 +660,154 @@ cc_sharing_panel_setup_media_sharing_dialog (CcSharingPanel *self)
...
+                                                                                NULL, &error);
+  if (error != NULL)
+    {
+      g_critical ("Unable to create a DBus proxy for the companion app avahi helper: %s", error->message);
+      g_error_free (error);
+
+      /* Return early, since there's no way we can set up the dialog from here */
+      return;
+    }
+  else
```
cosimoc: You can remove this else part.

panels/sharing/cc-sharing-panel.c
```diff
@@ -646,6 +660,154 @@ cc_sharing_panel_setup_media_sharing_dialog (CcSharingPanel *self)
...
+static void
+cc_sharing_panel_setup_content_sharing_dialog (CcSharingPanel *self)
+{
+  CcSharingPanelPrivate *priv = self->priv;
+  gboolean content_sharing_enabled;
+  GError *error = NULL;
+  GVariant *value;
+  GtkWidget *content_sharing_button = GTK_WIDGET (WID ("content-sharing-button"));
+
+  self->priv->companion_app_avahi_helper_proxy = g_dbus_proxy_new_for_bus_sync (G_BUS_TYPE_SYSTEM,
```
cosimoc: As mentioned in my previous review, please move this into `cc_sharing_panel_check_content_sharing_available()` and avoid showing the "Content sharing" list box row altogether when it fails.

panels/sharing/cc-sharing-panel.c
smspillaz: Ouch, yeah, good catch on the arguments.

My preference here would be to have a single atomic operation that sets up discoverability, and configures which name should be used (since we will have different names), rather than having multiple properties over the place. That is a personal design preference of mine, so I'm willing to override it if there's a good reason to have multiple properties.

panels/sharing/cc-sharing-panel.c
smspillaz: Hrm, I don't know if I would want to move it into *that* check, `cc_sharing_panel_check_content_sharing_available` is meant to be idempotent, whereas this operation here actually changes the program state by attempting to create the proxy.

In any event, if this fails, the row will not be shown, since the `content_sharing_button` widget is hidden by default:

```
                  <object class="GtkListBoxRow" id="content-sharing-button">
                    <property name="visible">False</property>
```

We only show it if everything succeeds:

```

  /* If everything was successful, show the row */
  gtk_widget_show (content_sharing_button);
  gtk_switch_set_active (GTK_SWITCH (self->priv->content_sharing_switch),
                         content_sharing_enabled);
  g_signal_connect (self->priv->content_sharing_switch, "notify::active",
                    G_CALLBACK (content_sharing_switch_active_changed_cb), self);
```

I think what made this confusing was this leftover:

```
  if (cc_sharing_panel_check_content_sharing_available ())
    cc_sharing_panel_setup_content_sharing_dialog (self);
  else
    gtk_widget_hide (WID ("content-sharing-button"));
```

Which would make the reader think that it was visible by default and needed to be hidden. I'll remove that.

panels/sharing/cc-sharing-panel.c
smspillaz: Done

panels/sharing/cc-sharing-panel.c
smspillaz: Done below

panels/sharing/cc-sharing-panel.c
cosimoc: I'm going to merge this as-is and offer a suggestion for a better approach in T20668.

PR #109, opened 2018-01-09
T20668

smspillaz:
https://phabricator.endlessm.com/T20668

### DIFF
--- a/panels/sharing/cc-sharing-panel.c
+++ b/panels/sharing/cc-sharing-panel.c
@@ -661,34 +661,70 @@ cc_sharing_panel_setup_media_sharing_dialog (CcSharingPanel *self)
 }
 
 static void
-content_sharing_switch_active_changed_cb (GtkSwitch *widget,
-                                          GParamSpec *pspec,
-                                          CcSharingPanel *self)
+enter_discoverable_mode_done_callback (GObject      *proxy,
+                                       GAsyncResult *result,
+                                       gpointer      user_data)
 {
+  CcSharingPanel *self = user_data;
   GError *error = NULL;
+  GVariant *return_tuple = g_dbus_proxy_call_finish (G_DBUS_PROXY (proxy), result, &error);
 
-  if (gtk_switch_get_active (widget))
+  if (!return_tuple)
     {
-      g_dbus_proxy_call_sync (self->priv->companion_app_avahi_helper_proxy,
-                              "EnterDiscoverableMode",
-                              g_variant_new ("(s)", "Endless Computer"),
-                              G_DBUS_CALL_FLAGS_NONE, -1,
-                              NULL, &error);
+      g_critical ("Error when calling EnterDiscoverableMode on Companion App Helper Proxy: %s",
+                  error->message);
+      g_clear_error (&error);
+      return;
     }
+
+  g_variant_unref (return_tuple);
+  gtk_switch_set_state (GTK_SWITCH (self->priv->content_sharing_switch), TRUE);
+}
+
+static void
+exit_discoverable_mode_done_callback (GObject      *proxy,
+                                      GAsyncResult *result,
+                                      gpointer      user_data)
+{
+  CcSharingPanel *self = user_data;
+  GError *error = NULL;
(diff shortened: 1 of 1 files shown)


panels/sharing/cc-sharing-panel.c
```diff
@@ -660,6 +660,31 @@ cc_sharing_panel_setup_media_sharing_dialog (CcSharingPanel *self)
                                            WID ("media-sharing-status-label"));
 }
 
+#define DMI_PRODUCT_NAME "/sys/class/dmi/id/product_name"
```
erikos: Maybe product_family is more appropriate here.

panels/sharing/cc-sharing-panel.c
```diff
@@ -660,6 +660,31 @@ cc_sharing_panel_setup_media_sharing_dialog (CcSharingPanel *self)
...
+
+/* We don't use the "Computer Name" here since that is really the hostname
+ * which is more of an internal detail. */
+static gchar *
+get_display_computer_name (void)
```
cosimoc: Instead of the product name, which is typically garbage, we should use the hostname, since it's also editable from the Details panel.
There's a nice dbus API for that; I think it should be possible to use it directly from the python script that writes the service file, so that we don't need to pass the name parameter over through the API.

At that point, the `Discoverable` property could be made read-write, since it would not need the name anymore, whose state is held separately by `hostnamectl`.

smspillaz: I suppose at the time we enable the companion app services we would want to read the hostname. Is that right?

cosimoc: Yeah, that sounds right; I was thinking you could read it directly from the python script that writes the service file

smspillaz:
Approach revised in light of the discussion we had about this yesterday.

panels/sharing/cc-sharing-panel.c
```diff
@@ -661,34 +661,69 @@ cc_sharing_panel_setup_media_sharing_dialog (CcSharingPanel *self)
...
+      g_clear_error (&error);
+      return;
     }
+
+  g_variant_unref (return_tuple);
+  gtk_switch_set_state (GTK_SWITCH (self->priv->content_sharing_switch), FALSE);
+}
+
+
+static void
```
cosimoc: The `state-set` callback has a `gboolean` return parameter, and this function should return `TRUE` to use the delayed state change machinery.

panels/sharing/cc-sharing-panel.c
```diff
@@ -661,34 +661,69 @@ cc_sharing_panel_setup_media_sharing_dialog (CcSharingPanel *self)
...
+      g_message ("Error when calling ExiDiscoverableMode on Companion App Helper Proxy: %s",
+                 error->message);
+      g_clear_error (&error);
+      return;
     }
+
+  g_variant_unref (return_tuple);
+  gtk_switch_set_state (GTK_SWITCH (self->priv->content_sharing_switch), FALSE);
+}
+
```
cosimoc: Remove empty line

panels/sharing/cc-sharing-panel.c
```diff
@@ -661,34 +661,69 @@ cc_sharing_panel_setup_media_sharing_dialog (CcSharingPanel *self)
...
+{
+  CcSharingPanel *self = user_data;
+  GError *error = NULL;
+  GVariant *return_tuple = g_dbus_proxy_call_finish (G_DBUS_PROXY (proxy), result, &error);
+
+  if (!return_tuple)
     {
-      g_critical ("Unable to set the enabled state of content sharing: %s", error->message);
-      g_error_free (error);
+      g_message ("Error when calling ExiDiscoverableMode on Companion App Helper Proxy: %s",
```
cosimoc: These should be critical messages

panels/sharing/cc-sharing-panel.c
smspillaz: Of course, I forgot about that.

panels/sharing/cc-sharing-panel.c
smspillaz: Also a typo :hear_no_evil:

PR #114, opened 2018-01-30
updates: Introduce Automatic Updates panel

GeorgesStavracas:
**Do not merge this yet; only for review until the rebase is finished.**

The Automatic Updates panel is a new panel that manages two
different settings:

 1. Whether updates should happen automatically or not
 2. A period where updates are allowed to happen

The first setting is implemented on a per-connection fashion,
using NetworkManager's ability to store user data on specific
connections. In this case, the property being set is:

  connection.allow-downloads-when-metered

The second setting is slightly trickier to implement, and uses
libmogwai-tariff to load and store a simple tariff. This tariff
has the following layout:

  i. Base tariff, downloads are forbidden, spans from 0 to forever.
 ii. Recurrent tariff, downloads are completely allowed, spans
     from [start hour] to [end hour]. If [start hour] >= [end hour],
     then [end hour] is contained in the following day.

This tariff schedule, when changed, is saved every 2 seconds, to
avoid abusing I/O.

https://phabricator.endlessm.com/T20818

### DIFF
--- a/panels/updates/Makefile.am
+++ b/panels/updates/Makefile.am
@@ -1,0 +1,42 @@
+# This is used in PANEL_CFLAGS
+cappletname = updates
+
+AM_CPPFLAGS = 						\
+	$(PANEL_CFLAGS)					\
+	$(UPDATES_PANEL_CFLAGS)				\
+	-DGNOMELOCALEDIR="\"$(datadir)/locale\""	\
+	$(NULL)
+
+BUILT_SOURCES =			\
+	cc-updates-resources.c	\
+	cc-updates-resources.h
+
+noinst_LTLIBRARIES = libupdates.la
+
+nodist_libupdates_la_SOURCES =	\
+	$(BUILT_SOURCES)
+
+libupdates_la_SOURCES =	\
+	cc-tariff-editor.c		\
+	cc-tariff-editor.h		\
+	cc-updates-panel.c		\
+	cc-updates-panel.h
+
+libupdates_la_LIBADD = $(PANEL_LIBS) $(UPDATES_PANEL_LIBS)
+
+resource_files = $(shell glib-compile-resources --sourcedir=$(srcdir) --generate-dependencies $(srcdir)/updates.gresource.xml)
+cc-updates-resources.c: updates.gresource.xml $(resource_files)
+	$(AM_V_GEN) glib-compile-resources --target=$@ --sourcedir=$(srcdir) --generate-source --c-name cc_updates $<
+cc-updates-resources.h: updates.gresource.xml $(resource_files)
+	$(AM_V_GEN) glib-compile-resources --target=$@ --sourcedir=$(srcdir) --generate-header --c-name cc_updates $<
+
+@INTLTOOL_DESKTOP_RULE@
+
+desktopdir = $(datadir)/applications
+desktop_in_files = gnome-updates-panel.desktop.in
+desktop_DATA = $(desktop_in_files:.desktop.in=.desktop)
+
+CLEANFILES = $(desktop_in_files) $(desktop_DATA) $(BUILT_SOURCES)
+EXTRA_DIST = $(resource_files) updates.gresource.xml
+
+-include $(top_srcdir)/git.mk
--- a/panels/updates/cc-tariff-editor.c
+++ b/panels/updates/cc-tariff-editor.c
@@ -1,0 +1,487 @@
+/* cc-tariff-editor.c
+ *
+ * Copyright © 2018 Georges Basile Stavracas Neto <georges.stavracas@gmail.com>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "cc-tariff-editor.h"
+
+#include <gdesktop-enums.h>
+#include <libmogwai-tariff/tariff-builder.h>
+
+#define CLOCK_SCHEMA     "org.gnome.desktop.interface"
+#define CLOCK_FORMAT_KEY "clock-format"
+#define SECONDS_PER_DAY  (24 * 60 * 60)
+
+struct _CcTariffEditor
+{
+  GtkGrid             parent;
+
+  GtkAdjustment      *adjustment_from_hours;
+  GtkAdjustment      *adjustment_from_minutes;
+  GtkAdjustment      *adjustment_to_hours;
+  GtkAdjustment      *adjustment_to_minutes;
+
+  GtkWidget          *stack_from;
+  GtkWidget          *stack_to;
+
+  GBytes             *tariff_as_bytes;
+
+  /* Clock format */
+  GSettings          *settings_clock;
+  GDesktopClockFormat clock_format;
+
+  guint8              time_period_from;
+  guint8              time_period_to;
+
+  guint64             seconds_to_start;
+  guint64             seconds_to_end;
+};
+
+G_DEFINE_TYPE (CcTariffEditor, cc_tariff_editor, GTK_TYPE_GRID)
+
+G_DEFINE_QUARK (CcTariffEditorError, cc_tariff_editor_error)
+
+enum
+{
+  TARIFF_CHANGED,
+  LAST_SIGNAL
+};
+
+enum
+{
+  AM,
+  PM,
+};
+
+
+static void on_time_changed_cb (CcTariffEditor *self);
+
+
+static guint signals[LAST_SIGNAL] = { 0, };
+
+
+/*
+ * Auxiliary methods
+ */
+
+static void
+update_seconds_from_adjustments (CcTariffEditor *self)
+{
+  guint factor_from = 0;
+  guint factor_to = 0;
+
+  if (self->clock_format == G_DESKTOP_CLOCK_FORMAT_12H)
+    {
+      factor_from = self->time_period_from == AM ? 0 : SECONDS_PER_DAY / 2;
+      factor_to = self->time_period_to == AM ? 0 : SECONDS_PER_DAY / 2;
+    }
+
+  self->seconds_to_start = gtk_adjustment_get_value (self->adjustment_from_hours) * 60 * 60 + factor_from +
+                           gtk_adjustment_get_value (self->adjustment_from_minutes) * 60;
+
+  self->seconds_to_end = gtk_adjustment_get_value (self->adjustment_to_hours) * 60 * 60 + factor_to +
+                         gtk_adjustment_get_value (self->adjustment_to_minutes) * 60;
+
+  /*
+   * If 'from' > 'to', e.g. [22:15, 05:45), this period is crossing days and we
+   * need to add +1 day to the end date. Otherwise, e.g, [13:10, 19:15), the period
+   * is contained in a single day, and we don't need to do anything.
+   */
+  if (self->seconds_to_start >= self->seconds_to_end)
+    self->seconds_to_end += SECONDS_PER_DAY;
+}
+
+static gboolean
+setup_tariff (CcTariffEditor  *self,
+              MwtTariff       *tariff,
+              GError         **error)
+{
+  g_autoptr(GDateTime) maximum_date = NULL;
+  GDateTime *period_start, *period_end;
+  MwtPeriod *base_period, *period;
+  GPtrArray *periods;
+
+  if (!tariff)
+    return FALSE;
+
+  /*
+   * Right now, it only parsed the same kind of tariff it generates, a tariff
+   * with these 2 periods:
+   *
+   *  1 → forbidden downloads period
+   *  2 → allowed downloads period
+   *
+   * Anything that does not comform to that is ignored.
+   */
+
+  periods = mwt_tariff_get_periods (tariff);
+
+  if (periods->len != 2)
+    {
+      g_set_error (error,
+                   CC_TARIFF_EDITOR_ERROR,
+                   CC_TARIFF_EDITOR_ERROR_NOT_SUPPORTED,
+                   "Cannot parse complex tariffs yet");
+      return FALSE;
+    }
+
+  /* Validate and setup the forbidden downloads period */
+  maximum_date = g_date_time_new_utc (9999, 12, 31, 23 , 59, 59);
+  base_period = g_ptr_array_index (periods, 0);
+  period = g_ptr_array_index (periods, 1);
+
+  if (mwt_period_get_repeat_type (base_period) != MWT_PERIOD_REPEAT_NONE ||
+      g_date_time_to_unix (mwt_period_get_start (base_period)) != 0 ||
+      g_date_time_to_unix (mwt_period_get_end (base_period)) != g_date_time_to_unix (maximum_date))
+    {
+      g_set_error (error,
+                   CC_TARIFF_EDITOR_ERROR,
+                   CC_TARIFF_EDITOR_ERROR_WRONG_FORMAT,
+                   "Base tariff is wrong");
+      return FALSE;
+    }
+
+  if (mwt_period_get_repeat_type (period) != MWT_PERIOD_REPEAT_DAY ||
+      mwt_period_get_repeat_period (period) != 1)
+    {
+      g_set_error (error,
+                   CC_TARIFF_EDITOR_ERROR,
+                   CC_TARIFF_EDITOR_ERROR_WRONG_FORMAT,
+                   "Repeating period is wrong");
+      return FALSE;
+    }
+
+  g_signal_handlers_block_by_func (self->adjustment_from_hours, on_time_changed_cb, self);
+  g_signal_handlers_block_by_func (self->adjustment_from_minutes, on_time_changed_cb, self);
+  g_signal_handlers_block_by_func (self->adjustment_to_hours, on_time_changed_cb, self);
+  g_signal_handlers_block_by_func (self->adjustment_to_minutes, on_time_changed_cb, self);
+
+  period_start = mwt_period_get_start (period);
+  period_end = mwt_period_get_end (period);
+
+  if (self->clock_format == G_DESKTOP_CLOCK_FORMAT_24H)
+    {
+      gtk_adjustment_set_value (self->adjustment_from_hours, g_date_time_get_hour (period_start));
+      gtk_adjustment_set_value (self->adjustment_from_minutes, g_date_time_get_minute (period_start));
+      gtk_adjustment_set_value (self->adjustment_to_hours, g_date_time_get_hour (period_end));
+      gtk_adjustment_set_value (self->adjustment_to_minutes, g_date_time_get_minute (period_end));
+    }
+  else
+    {
+      self->time_period_from = g_date_time_get_hour (period_start) < 12 ? AM : PM;
+      self->time_period_to = g_date_time_get_hour (period_end) < 12 ? AM : PM;
+
+      gtk_stack_set_visible_child_name (GTK_STACK (self->stack_from), self->time_period_from == AM ? "am" : "pm");
+      gtk_stack_set_visible_child_name (GTK_STACK (self->stack_to), self->time_period_to == AM ? "am" : "pm");
+
+      /* Special-case: 12 AM/PM must show up as 12, not as 0 AM/PM */
+      if (g_date_time_get_hour (period_start) % 12 == 0)
+        gtk_adjustment_set_value (self->adjustment_from_hours, 12);
+      else
+        gtk_adjustment_set_value (self->adjustment_from_hours, g_date_time_get_hour (period_start) % 12);
+
+      if (g_date_time_get_hour (period_end) % 12 == 0)
+        gtk_adjustment_set_value (self->adjustment_to_hours, 12);
+      else
+        gtk_adjustment_set_value (self->adjustment_to_hours, g_date_time_get_hour (period_end) % 12);
+
+      gtk_adjustment_set_value (self->adjustment_from_minutes, g_date_time_get_minute (period_start));
+      gtk_adjustment_set_value (self->adjustment_to_minutes, g_date_time_get_minute (period_end));
+    }
+
+  g_signal_handlers_unblock_by_func (self->adjustment_from_hours, on_time_changed_cb, self);
+  g_signal_handlers_unblock_by_func (self->adjustment_from_minutes, on_time_changed_cb, self);
+  g_signal_handlers_unblock_by_func (self->adjustment_to_hours, on_time_changed_cb, self);
+  g_signal_handlers_unblock_by_func (self->adjustment_to_minutes, on_time_changed_cb, self);
+
+  return TRUE;
+}
+
+static void
+update_tariff (CcTariffEditor *self,
+               gboolean        should_notify)
+{
+  g_autoptr(MwtTariffBuilder) tariff_builder = NULL;
+  g_autoptr(MwtPeriod) forbidden_period = NULL;
+  g_autoptr(GDateTime) forbidden_start = NULL;
+  g_autoptr(GDateTime) forbidden_end = NULL;
+  g_autoptr(MwtPeriod) allowed_period = NULL;
+  g_autoptr(GDateTime) allowed_start = NULL;
+  g_autoptr(GDateTime) allowed_end = NULL;
+
+  /*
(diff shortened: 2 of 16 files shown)


panels/updates/cc-updates-panel.c
```diff
@@ -0,0 +1,688 @@
...
+
+  tariff_file = g_file_new_for_path (tariff_path);
+  input_stream = G_INPUT_STREAM (g_file_read (tariff_file, NULL, &error));
+  if (error)
+    {
+      g_warning ("Error reading test tariff file: %s", error->message);
+      return;
+    }
+
+  bytes = g_input_stream_read_bytes (input_stream, 8192, NULL, &error);
```
pwithnall: This should be async. Or, if you don’t want it to be async, just use `g_file_get_contents()` and `g_bytes_new_take()`.

panels/updates/cc-updates-panel.c
```diff
@@ -0,0 +1,688 @@
...
+
+  tariff_path = g_build_filename (g_get_user_config_dir (), SYSTEM_TARIFF_FILENAME, NULL);
+
+  if (!g_file_test (tariff_path, G_FILE_TEST_EXISTS))
+    {
+      g_debug ("Could not open test tariff file");
+      return;
+    }
+
+  if (!g_file_test (tariff_path, G_FILE_TEST_IS_REGULAR))
```
pwithnall: No need for these explicit `g_file_test()` calls: just handle the appropriate errors from `g_file_read()`. Otherwise you have a [TOCTTOU race](https://en.wikipedia.org/wiki/Time_of_check_to_time_of_use).

panels/updates/cc-updates-panel.c
```diff
@@ -0,0 +1,688 @@
...
+  if (error)
+    {
+      g_warning ("Error loading test tariff: %s", error->message);
+    }
+  else
+    {
+      gtk_switch_set_active (GTK_SWITCH (self->scheduled_updates_switch), TRUE);
+    }
+
+
```
pwithnall: Nitpick: Spurious double empty line.

panels/updates/cc-updates-panel.c
```diff
@@ -0,0 +1,688 @@
...
+  g_signal_handlers_unblock_by_func (self->scheduled_updates_switch, on_scheduled_updates_switch_changed_cb, self);
+  g_signal_handlers_unblock_by_func (self->tariff_editor, on_tariff_changed_cb, self);
+
+  g_input_stream_close (input_stream, NULL, &error);
+  if (error)
+    g_warning ("Error closing test tariff file: %s", error->message);
```
pwithnall: s/test//?

panels/updates/cc-updates-panel.c
```diff
@@ -0,0 +1,688 @@
...
+  g_signal_handlers_block_by_func (self->tariff_editor, on_tariff_changed_cb, self);
+  g_signal_handlers_block_by_func (self->scheduled_updates_switch, on_scheduled_updates_switch_changed_cb, self);
+
+  cc_tariff_editor_load_tariff (self->tariff_editor,
+                                mwt_tariff_loader_get_tariff (tariff_loader),
+                                &error);
+
+  if (error)
+    {
+      g_warning ("Error loading test tariff: %s", error->message);
```
pwithnall: s/test//?

panels/updates/cc-updates-panel.c
```diff
@@ -0,0 +1,688 @@
...
+    {
+      g_warning ("Error reading test tariff file: %s", error->message);
+      return;
+    }
+
+  tariff_loader = mwt_tariff_loader_new ();
+  mwt_tariff_loader_load_from_bytes (tariff_loader, bytes, &error);
+  if (error)
+    {
+      g_warning ("Error loading test tariff: %s", error->message);
```
pwithnall: s/test//?

panels/updates/cc-updates-panel.c
```diff
@@ -0,0 +1,688 @@
...
+  if (error)
+    {
+      g_warning ("Error reading test tariff file: %s", error->message);
+      return;
+    }
+
+  bytes = g_input_stream_read_bytes (input_stream, 8192, NULL, &error);
+  if (error)
+    {
+      g_warning ("Error reading test tariff file: %s", error->message);
```
pwithnall: s/test//?

panels/updates/cc-updates-panel.c
```diff
@@ -0,0 +1,688 @@
...
+    {
+      g_debug ("Test tariff file is not a regular file");
+      return;
+    }
+
+  tariff_file = g_file_new_for_path (tariff_path);
+  input_stream = G_INPUT_STREAM (g_file_read (tariff_file, NULL, &error));
+  if (error)
+    {
+      g_warning ("Error reading test tariff file: %s", error->message);
```
pwithnall: s/test//?

panels/updates/cc-updates-panel.c
```diff
@@ -0,0 +1,688 @@
...
+  g_autofree gchar *tariff_path = NULL;
+  GBytes *tariff_bytes;
+
+  tariff_path = g_build_filename (g_get_user_config_dir (), SYSTEM_TARIFF_FILENAME, NULL);
+  tariff_bytes = cc_tariff_editor_get_tariff_as_bytes (self->tariff_editor);
+
+  if (!tariff_bytes || !gtk_switch_get_active (GTK_SWITCH (self->scheduled_updates_switch)))
+    {
+      /* Remove the tariff file if NULL */
+      if (g_file_test (tariff_path, G_FILE_TEST_EXISTS))
```
pwithnall: Again, using `g_file_test()` introduces a TOCTTOU race. You can just check the error value from `g_file_delete()`. (You would have to handle errors from `g_file_delete()` anyway.)

panels/updates/cc-updates-panel.c
```diff
@@ -0,0 +1,688 @@
...
+        {
+          g_autoptr(GFile) file = g_file_new_for_path (tariff_path);
+
+          g_debug ("Deleting tariff");
+
+          g_file_delete (file, NULL, &error);
```
pwithnall: This should be async.

panels/updates/cc-updates-panel.c
```diff
@@ -0,0 +1,688 @@
...
+static void
+cc_updates_panel_finalize (GObject *object)
+{
+  CcUpdatesPanel *self = (CcUpdatesPanel *)object;
+
+  g_cancellable_cancel (self->cancellable);
+  g_clear_object (&self->cancellable);
+  g_clear_object (&self->current_device);
+  g_clear_object (&self->nm_client);
+
```
pwithnall: Need to disconnect `changed_id` too.

panels/updates/cc-updates-panel.c
```diff
@@ -0,0 +1,688 @@
...
+  /* "Change Network Settings" label */
+  settings_text = g_strdup_printf ("<a href=\"\">%s</a>", _("Change Network Settings…"));
+  gtk_label_set_markup (GTK_LABEL (self->network_settings_label), settings_text);
+
+  /* Load any saved tariff */
+  load_tariff (self);
+
+  /* Setup network manager */
+  self->nm_client = nm_client_new (NULL, NULL);
```
pwithnall: This should be async. It creates a D-Bus proxy to connect to NetworkManager.

panels/updates/cc-updates-panel.c
pwithnall: Actually, I was thinking. `mogwai-scheduled` needs to be able to read the tariff for each connection in order to know when to allow downloads. So either it needs access to `~/.config/system-tariff` (which is not a good idea from a sandboxing point of view: a system daemon with access to a user’s home directory; it also doesn’t work if the connection is shared between multiple users by NM), or the tariffs need to be stored by `mogwai-scheduled`. I don’t particularly fancy writing multi-user storage code in `mogwai-scheduled`, but then I wondered:

We could just store the tariff on the `NMSettingUser`, in addition to the property saying whether to download when the connection is metered. The `NMSettingUser` code requires that settings values are 8KiB or smaller, and valid UTF-8 (see `nm_setting_user_check_val()`). That should be fine for storing a tariff, if you take the `GVariant` format of the tariff, and convert it to/from UTF-8 using `g_variant_print()` and `g_variant_parse()`.

That should simplify the async handling here too, since you’d no longer be dealing with files; just with `NMSettingUser`. What do you think?

panels/updates/cc-updates-panel.ui
```diff
@@ -0,0 +1,324 @@
...
+                          </object>
+                        </child>
+
+                        <child>
+                          <object class="GtkLabel">
+                            <property name="visible">True</property>
+                            <property name="can_focus">False</property>
+                            <property name="hexpand">True</property>
+                            <property name="xalign">0.0</property>
+                            <property name="label" translatable="yes">Allow system and application updates to happen automatically without your request in background.</property>
```
pwithnall: Nitpick: s/in background/in the background/

panels/updates/cc-updates-panel.c
pwithnall: One complication here is that, currently, the UI assumes the tariff applies to all NM connections. The approach I’m suggesting applies it only to the currently active connection. Which do we want? Eventually, presumably, the UI will need to allow different tariffs to be configured for different connections; what compromise solution would be best here?

panels/updates/cc-updates-panel.c
GeorgesStavracas: At first, I did that. But appearently NetworkManager destroys all the `NMDevices` and we end up with an invalid source id at this point. Disconnecting in `dispose()` also yields the same results.

panels/updates/cc-updates-panel.c
GeorgesStavracas: It is common (but not good) practice in Settings to load this synchronously. I am going to rework that upstream for 3.28, and since by then we'd have to rework this initialization code anyway, I concluded it's not worth the hassle.

panels/updates/cc-tariff-editor.c
```diff
@@ -0,0 +1,485 @@
...
+/*
+ * GObject overrides
+ */
+
+static void
+cc_tariff_editor_finalize (GObject *object)
+{
+  CcTariffEditor *self = (CcTariffEditor *)object;
+
+  g_clear_object (&self->settings_clock);
```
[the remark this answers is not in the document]
GeorgesStavracas: Good catch

panels/updates/cc-tariff-editor.c
```diff
@@ -0,0 +1,485 @@
...
+        gtk_adjustment_set_value (self->adjustment_from_hours, g_date_time_get_hour (period_start) % 12);
+
+      if (g_date_time_get_hour (period_end) % 12 == 0)
+        gtk_adjustment_set_value (self->adjustment_to_hours, 12);
+      else
+        gtk_adjustment_set_value (self->adjustment_to_hours, g_date_time_get_hour (period_end) % 12);
+
+      gtk_adjustment_set_value (self->adjustment_from_minutes, g_date_time_get_minute (period_start));
+      gtk_adjustment_set_value (self->adjustment_to_minutes, g_date_time_get_minute (period_end));
+    }
```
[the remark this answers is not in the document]
GeorgesStavracas: Improved the comment above it.

panels/updates/Makefile.am
```diff
@@ -0,0 +1,40 @@
...
+cc-updates-resources.h: updates.gresource.xml $(resource_files)
+	$(AM_V_GEN) glib-compile-resources --target=$@ --sourcedir=$(srcdir) --generate-header --c-name cc_updates $<
+
+@INTLTOOL_DESKTOP_RULE@
+
+desktopdir = $(datadir)/applications
+desktop_in_files = gnome-updates-panel.desktop.in
+desktop_DATA = $(desktop_in_files:.desktop.in=.desktop)
+
+CLEANFILES = $(Desktop_in_files) $(desktop_DATA) $(BUILT_SOURCES)
```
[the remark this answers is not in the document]
pwithnall: Ah yes, the `.desktop.in.in` should be distributed as a result of its listing in `AC_CONFIG_FILES`.

panels/updates/cc-updates-panel.c
```diff
@@ -0,0 +1,688 @@
...
+  if (self->changed_id > 0)
+    {
+      g_source_remove (self->changed_id);
+      self->changed_id = 0;
+    }
+
+  /* Setup the new device */
+  if (g_set_object (&self->current_device, device) && device)
+    {
+      self->changed_id = g_signal_connect_swapped (device,
```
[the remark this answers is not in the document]
pwithnall: This is still an issue. Did you mean to use `g_signal_handler_disconnect()` instead of `g_source_remove()`?

panels/updates/cc-updates-panel.c
```diff
@@ -0,0 +1,688 @@
...
+  icon_name = get_network_status_icon_name (self, device, ap);
+  gtk_image_set_from_icon_name (GTK_IMAGE (self->network_status_icon), icon_name, 4);
+
+  /* Name */
+  if (ap)
+    {
+      GBytes *ssid = nm_access_point_get_ssid (ap);
+      g_autofree gchar *title = nm_utils_ssid_to_utf8 (g_bytes_get_data (ssid, NULL), g_bytes_get_size (ssid));
+
+      gtk_label_set_label (GTK_LABEL (self->network_name_label), title);
```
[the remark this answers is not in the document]
pwithnall: Let’s leave it as it is for now. I guess the network ID isn’t completely malicious, because you would have previously chosen to connect to that network.

panels/updates/cc-updates-panel.c
```diff
@@ -0,0 +1,684 @@
...
+}
+
+static void
+load_tariff (CcUpdatesPanel *self)
+{
+  g_autoptr(MwtTariffLoader) tariff_loader = NULL;
+  g_autoptr(GBytes) bytes = NULL;
+  g_autoptr(GError) error = NULL;
+  g_autofree gchar *tariff_path = NULL;
+  gchar *contents = NULL;
```
pwithnall: This should be `g_autofree`

panels/updates/cc-updates-panel.c
```diff
@@ -0,0 +1,684 @@
...
+  tariff_path = g_build_filename (g_get_user_config_dir (), SYSTEM_TARIFF_FILENAME, NULL);
+  g_file_get_contents (tariff_path, &contents, &size, &error);
+  if (error)
+    {
+      if (!g_error_matches (error, G_FILE_ERROR, G_FILE_ERROR_NOENT))
+        g_warning ("Error reading tariff file: %s", error->message);
+      return;
+    }
+
+  bytes = g_bytes_new_take (contents, size);
```
pwithnall: …and use `g_steal_pointer()` here. I realise this makes no functional difference, but it does make the ownership transfer more obvious, and means that refactoring the code is less likely to cause a double-free or leak.

panels/updates/cc-updates-panel.c
```diff
@@ -0,0 +1,684 @@
...
+}
+
+static void
+save_tariff (CcUpdatesPanel *self)
+{
+  g_autoptr(GError) error = NULL;
+  g_autofree gchar *tariff_path = NULL;
+  GBytes *tariff_bytes;
+
+  tariff_path = g_build_filename (g_get_user_config_dir (), SYSTEM_TARIFF_FILENAME, NULL);
```
pwithnall: Nitpick: You could factor out the `g_build_filename()` calls from this and the previous function, into a singe `get_tariff_filename()` helper.

panels/updates/cc-updates-panel.c
```diff
@@ -0,0 +1,684 @@
...
+                           G_PRIORITY_LOW,
+                           NULL,
+                           (GAsyncReadyCallback) on_tariff_file_deleted_cb,
+                           &error);
+      return;
+    }
+
+  g_debug ("Saving tariff");
+
+  g_file_set_contents (tariff_path,
```
pwithnall: If deleting the file is async, then setting the file’s contents should probably also be async.

panels/updates/cc-updates-panel.c
```diff
@@ -0,0 +1,684 @@
...
+static void
+on_network_changed_cb (CcUpdatesPanel *self)
+{
+  g_debug ("NetworkManager changed state");
+
+  update_active_network (self);
+}
+
+static void
+on_network_changes_commited_cb (GObject      *source,
```
pwithnall: Nitpick: s/commited/committed/

panels/updates/cc-updates-panel.c
```diff
@@ -0,0 +1,684 @@
...
+    {
+      g_autoptr(GFile) file = g_file_new_for_path (tariff_path);
+
+      g_debug ("Deleting tariff");
+
+      g_file_delete_async (file,
+                           G_PRIORITY_LOW,
+                           NULL,
+                           (GAsyncReadyCallback) on_tariff_file_deleted_cb,
+                           &error);
```
pwithnall: `&error` is being passed as the `user_data` here, but `on_tariff_file_deleted_cb()` assumes that the `user_data` is `self`.

panels/updates/cc-updates-panel.c
```diff
@@ -0,0 +1,684 @@
...
+    {
+      g_warning ("Error storing "NM_SETTING_ALLOW_DOWNLOADS_WHEN_METERED": %s", error->message);
+      return;
+    }
+
+  nm_remote_connection_commit_changes_async (NM_REMOTE_CONNECTION (connection),
+                                             TRUE, /* save to disk */
+                                             NULL,
+                                             on_network_changes_commited_cb,
+                                             NULL);
```
pwithnall: This async call doesn’t hold a reference to `self`, so it’s possible that the `CcUpdatesPanel` will be destroyed before the async operation finishes. This is a bit tricky to fix. There are two possible solutions:
 * Hold a strong reference to the `CcUpdatesPanel`, which means that a slow async operation can keep the panel alive longer than expected.
 * Hold a weak reference, and pass in a `GCancellable` to the async call. When the `CcUpdatesPanel` is disposed, cancel the `GCancellable` so that the async operation completes.

panels/updates/cc-updates-panel.c
```diff
@@ -0,0 +1,688 @@
...
+  g_autoptr(MwtTariffLoader) tariff_loader = NULL;
+  g_autoptr(GInputStream) input_stream = NULL;
+  g_autoptr(GBytes) bytes = NULL;
+  g_autoptr(GFile) tariff_file = NULL;
+  g_autoptr(GError) error = NULL;
+  g_autofree gchar *tariff_path = NULL;
+
+  g_debug ("Loading tariff");
+
+  tariff_path = g_build_filename (g_get_user_config_dir (), SYSTEM_TARIFF_FILENAME, NULL);
```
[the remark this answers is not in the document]
pwithnall: Are you planning on changing this, or do you think it’s not worth it?

panels/updates/cc-updates-panel.c
pwithnall: That doesn’t make sense, since you’re holding a strong reference to the device in `current_device`, right? As far as I understand it, `changed_id` is a signal handler ID, not a source ID, though. Were you using `g_source_remove()` here? That wouldn’t be valid; use `g_signal_handler_disconnect()`.

panels/updates/cc-updates-panel.c
pwithnall: OK. Please leave a `/* FIXME: Rework this to be async. This will be done at the same time as reworking all the other panels to be async. */` comment there to make that obvious.

panels/updates/cc-updates-panel.c
GeorgesStavracas: /me facepalms
Please pretend we never had this discussion.

panels/updates/cc-updates-panel.c
GeorgesStavracas: The callback does not access anything from `CcUpdatesPanel`. I don't understand why it is necessary to keep the panel alive in this case.

panels/updates/cc-updates-panel.c
GeorgesStavracas: The `&error` was obviously a typo, but it's the same case from before: the `CcUpdatesPanel` is not accessed in the callback.

panels/updates/cc-updates-panel.c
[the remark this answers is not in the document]
GeorgesStavracas: Sure.

panels/updates/cc-updates-panel.c
GeorgesStavracas: At this point, we're dealing with a tariff file that is exactly 110 bytes long. I'm afraid the added complexity of asynchronously saving the tariff surpasses the benefits.

panels/updates/cc-updates-panel.c
pwithnall: If, for example, the user changes the tariff, then immediately closes g-c-c, the following could happen:
 * Saving the changes starts
 * The async start function returns, and control goes back to the main context
 * The main loop is quit (I suspect that closing the g-c-c window will end up calling `g_main_loop_quit()` to achieve this, but haven’t checked)
 * The necessary main context iterations to finish the D-Bus communication necessary to finish saving the NetworkManager settings do not happen, and hence the settings potentially don’t get saved

In all likelihood, the D-Bus traffic here is a single message, and so the settings will be correctly saved, even if the user does somehow manage to hit the race described above. However, proving that the savings will always be saved properly is really hard.

What needs to happen is that the `GMainContext` continues to be iterated on shutdown of g-c-c until all pending operations (from all panels) are compete (after hiding the main window). However, I realise that’s a fairly major change.

For the moment, I’d be fine with a `FIXME` comment here mentioning that, and a plan to implement this blocking shutdown at the same time as making panel startup async (because that will expose more of the same kind of race).

panels/updates/cc-updates-panel.c
```diff
@@ -285,52 +319,100 @@ get_network_status_icon_name (CcUpdatesPanel *self,
...
-  /* Setup the new device */
-  if (g_set_object (&self->current_device, device) && device)
+  if (connection)
     {
-      /* Cleanup previous device handler */
-      if (self->changed_id > 0)
+      NMSettingUser *setting_user;
+      const gchar *tariff_enabled_value = NULL;
+
+      setting_user = NM_SETTING_USER (nm_connection_get_setting (connection, NM_TYPE_SETTING_USER));
```
pwithnall: `nm_connection_get_setting()` can apparently return `NULL`. Can that ever happen for `NM_TYPE_SETTING_USER`? If so, the null value needs to be handled gracefully. If not, please add a `g_assert (setting_user != NULL);`.

panels/updates/cc-updates-panel.c
```diff
@@ -285,52 +319,100 @@ get_network_status_icon_name (CcUpdatesPanel *self,
...
+
+  if (connection)
     {
-      GBytes *ssid = nm_access_point_get_ssid (ap);
-      g_autofree gchar *title = nm_utils_ssid_to_utf8 (g_bytes_get_data (ssid, NULL), g_bytes_get_size (ssid));
+      NMSettingUser *setting_user;
+      const gchar *value;
 
-      gtk_label_set_label (GTK_LABEL (self->network_name_label), title);
+      setting_user = NM_SETTING_USER (nm_connection_get_setting (connection, NM_TYPE_SETTING_USER));
```
pwithnall: `nm_connection_get_setting()` can potentially return `NULL` here too.

panels/updates/cc-updates-panel.c
```diff
@@ -346,119 +428,80 @@ update_active_network (CcUpdatesPanel *self)
...
+                                                   "state-changed",
+                                                   G_CALLBACK (on_network_changed_cb),
+                                                   self);
     }
 
-  g_signal_handlers_block_by_func (self->tariff_editor, on_tariff_changed_cb, self);
-  g_signal_handlers_block_by_func (self->scheduled_updates_switch, on_scheduled_updates_switch_changed_cb, self);
+  /* Icon */
+  icon_name = get_network_status_icon_name (self, device, ap);
+  gtk_image_set_from_icon_name (GTK_IMAGE (self->network_status_icon), icon_name, 4);
```
pwithnall: s/`4`/`GTK_ICON_SIZE_BUTTON`/? (I think that’s the right enum value; please check.)

panels/updates/cc-updates-panel.c
```diff
@@ -346,119 +428,80 @@ update_active_network (CcUpdatesPanel *self)
...
-    g_warning ("Error loading tariff: %s", error->message);
+      gtk_label_set_label (GTK_LABEL (self->network_name_label), title);
+    }
   else
-    gtk_switch_set_active (GTK_SWITCH (self->scheduled_updates_switch), TRUE);
+    {
+      if (connection)
+        gtk_label_set_label (GTK_LABEL (self->network_name_label), _("Connected"));
+      else
+        gtk_label_set_label (GTK_LABEL (self->network_name_label), _("No active connection"));
```
pwithnall: Nitpick: This could be refactored to only have a single `gtk_label_set_label()` call:
```
gtk_label_set_label (GTK_LABEL (self->network_name_label), connection ? _("Connected") : _("No active connection"));
```

pwithnall:
(New version in #116.)

PR #116, opened 2018-02-01
T20818

GeorgesStavracas:
**Blocked by: #115**

https://phabricator.endlessm.com/T20818

### DIFF
--- a/panels/updates/cc-updates-panel.c
+++ b/panels/updates/cc-updates-panel.c
@@ -1,0 +1,718 @@
+/* cc-updates-panel.c
+ *
+ * Copyright © 2018 Endless, Inc
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authors: Georges Basile Stavracas Neto <georges@endlessm.com>
+ */
+
+#include "cc-tariff-editor.h"
+#include "cc-updates-panel.h"
+#include "cc-updates-resources.h"
+
+#include <glib/gi18n.h>
+#include <libmogwai-tariff/tariff-loader.h>
+#include <NetworkManager.h>
+
+#define NM_SETTING_ALLOW_DOWNLOADS_WHEN_METERED "connection.allow-downloads-when-metered"
+#define NM_SETTING_TARIFF_ENABLED               "connection.tariff-enabled"
+#define NM_SETTING_TARIFF                       "connection.tariff"
+
+struct _CcUpdatesPanel
+{
+  CcPanel             parent;
+
+  GtkWidget          *automatic_updates_container;
+  GtkWidget          *automatic_updates_switch;
+  GtkWidget          *metered_data_label;
+  GtkWidget          *network_name_label;
+  GtkWidget          *network_settings_label;
+  GtkWidget          *network_status_icon;
+  GtkWidget          *scheduled_updates_container;
+  GtkWidget          *scheduled_updates_switch;
+  CcTariffEditor     *tariff_editor;
+
+  /* Network Manager */
+  NMClient           *nm_client;
+  NMDevice           *current_device;
+
+  /* Signal handlers */
+  gulong              changed_id;
+  guint               save_tariff_timeout_id;
+
+  GCancellable       *cancellable;
+};
+
+
+static void          on_automatic_updates_switch_changed_cb      (GtkSwitch      *sw,
+                                                                  GParamSpec     *pspec,
+                                                                  CcUpdatesPanel *self);
+
+static void          on_network_changed_cb                       (CcUpdatesPanel *self);
+
+static void          on_network_changes_committed_cb             (GObject        *source,
+                                                                  GAsyncResult   *result,
+                                                                  gpointer        user_data);
+
+static void          on_scheduled_updates_switch_changed_cb      (GtkSwitch      *sw,
+                                                                  GParamSpec     *pspec,
+                                                                  CcUpdatesPanel *self);
+
+static void          on_tariff_changed_cb                        (CcTariffEditor *tariff_editor,
+                                                                  CcUpdatesPanel *self);
+
+
+G_DEFINE_TYPE (CcUpdatesPanel, cc_updates_panel, CC_TYPE_PANEL)
+
+enum
+{
+  PROP_PARAMETERS = 1,
+  N_PROPS
+};
+
+
+/*
+ * Auxiliary methods
+ */
+
+static void
+cleanup_signals (CcUpdatesPanel *self)
+{
+  if (self->changed_id > 0)
+    {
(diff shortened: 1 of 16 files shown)


panels/updates/cc-updates-panel.c
```diff
@@ -0,0 +1,714 @@
...
+                                 gboolean      automatic_updates_enabled,
+                                 gboolean      tariff_enabled,
+                                 GVariant     *tariff_variant)
+{
+  NMSettingUser *setting_user;
+  g_autofree gchar *tariff_string = NULL;
+  g_autoptr(GError) error = NULL;
+
+  setting_user = NM_SETTING_USER (nm_connection_get_setting (connection, NM_TYPE_SETTING_USER));
+  g_assert_nonnull (setting_user);
```
pwithnall: Nitpick: `g_assert_nonnull()` should only be used for unit tests, since it’s not compiled out when compiling with `G_DISABLE_CHECKS`. `g_assert()` is (and hence you get improved performance with `G_DISABLE_CHECKS`). Use `g_assert()` in application/library code, and `g_assert_*()` in unit tests.

panels/updates/cc-updates-panel.c
```diff
@@ -0,0 +1,714 @@
...
+      gboolean enabled;
+
+      g_debug ("Creating a new custom config file for the current connection…");
+
+      /* Add a new NMSettingUser to this connection */
+      setting_user = NM_SETTING_USER (nm_setting_user_new ());
+      nm_connection_add_setting (connection, NM_SETTING (setting_user));
+
+      /* The default value depends on the metered state of the connection */
+      setting = nm_connection_get_setting_connection (connection);
```
pwithnall: I think `nm_connection_get_setting_connection()` can also potentially return `NULL`.

panels/updates/cc-updates-panel.c
```diff
@@ -0,0 +1,714 @@
...
+
+  g_debug ("Loading tariff from connection");
+
+  if (connection)
+    {
+      NMSettingUser *setting_user;
+      const gchar *tariff_enabled_value = NULL;
+
+      setting_user = NM_SETTING_USER (nm_connection_get_setting (connection, NM_TYPE_SETTING_USER));
+      g_assert_nonnull (setting_user);
```
pwithnall: `g_assert (setting_user != NULL);`

panels/updates/cc-updates-panel.c
```diff
@@ -0,0 +1,714 @@
...
+
+  g_signal_handlers_block_by_func (self->automatic_updates_switch, on_automatic_updates_switch_changed_cb, self);
+
+  if (connection)
+    {
+      NMSettingUser *setting_user;
+      const gchar *value;
+
+      setting_user = NM_SETTING_USER (nm_connection_get_setting (connection, NM_TYPE_SETTING_USER));
+      g_assert_nonnull (setting_user);
```
pwithnall: `g_assert (setting_user != NULL);`

panels/updates/cc-updates-panel.c
```diff
@@ -0,0 +1,714 @@
...
+                                    NMConnection   *connection)
+{
+  gtk_widget_set_visible (self->metered_data_label, connection != NULL);
+
+  if (connection)
+    {
+      NMSettingConnection *setting;
+      NMMetered metered;
+
+      setting = nm_connection_get_setting_connection (connection);
```
pwithnall: `nm_connection_get_setting_connection()` could potentially return `NULL` here.

panels/updates/cc-updates-panel.c
```diff
@@ -0,0 +1,732 @@
...
+    }
+  else
+    {
+      gtk_label_set_label (GTK_LABEL (self->network_name_label),
+                           connection ? _("Connected") : _("No active connection"));
+    }
+
+  /* This makes sure we never return NULL to 'nm_connection_get_setting (c, NM_TYPE_SETTING_USER)' */
+  ensure_setting_connection (self, connection);
+  ensure_setting_user (self, connection);
```
pwithnall: I’m not 100% convinced these calls are sufficient to ensure you always have a `NMSettingConnection` or `NMSettingUser` when you need it. I think it would be more obviously correct to change `ensure_setting_connection()` to have the same declaration as `nm_connection_get_setting_user()`, replace all calls to the latter with the former, and have the implementation of `ensure_setting_connection()` be something like:

```
static NMSettingConnection *
ensure_setting_connection (NMConnection *connection)
{
  NMSettingConnection *setting = nm_connection_get_setting_connection (connection);
  if (setting == NULL)
    {
      setting = NM_SETTING_CONNECTION (nm_setting_connection_new ());
      nm_connection_add_setting (connection, NM_SETTING (setting));
    }

  return setting;
}
```

Same for `ensure_setting_user()`.

panels/updates/cc-updates-panel.c
pwithnall: I’ve left a follow-up comment on the `ensure_setting_connection()` stuff.

panels/updates/cc-updates-panel.c
pwithnall: I’ve left a follow-up comment on the `ensure_setting_connection()` stuff.

panels/updates/cc-updates-panel.c
```diff
@@ -0,0 +1,732 @@
...
+                                 gboolean      automatic_updates_enabled,
+                                 gboolean      tariff_enabled,
+                                 GVariant     *tariff_variant)
+{
+  NMSettingUser *setting_user;
+  g_autofree gchar *tariff_string = NULL;
+  g_autoptr(GError) error = NULL;
+
+  setting_user = NM_SETTING_USER (nm_connection_get_setting (connection, NM_TYPE_SETTING_USER));
+  g_assert (setting_user != NULL);
```
pwithnall: You could drop these two lines and use `ensure_setting_user()` instead.

panels/updates/cc-updates-panel.c
```diff
@@ -0,0 +1,732 @@
...
+      NMMetered metered;
+      gboolean enabled;
+
+      g_debug ("Creating a new custom config file for the current connection…");
+
+      /* Add a new NMSettingUser to this connection */
+      setting_user = NM_SETTING_USER (nm_setting_user_new ());
+      nm_connection_add_setting (connection, NM_SETTING (setting_user));
+
+      /* The default value depends on the metered state of the connection */
```
pwithnall: I was thinking that `ensure_setting_user()` would just be the lines up to `nm_connection_add_setting()` (i.e. like `ensure_setting_connection()`). The lines below here which set up signal handlers and modify the UI should probably be moved elsewhere — probably to just after where you first call `ensure_setting_user()`.

GeorgesStavracas: I disagree. The new panel assumes that we always have a value set for those settings. I don't see what's the problem with those lines. I also see no point in splitting this part of code, for it is too related to what's happening there. If no NMSettingUser is set, set the default and save and return; otherwise, just return.

There are no signal handlers being set.

GeorgesStavracas: Also, the whole point of the assert there is that it is absolutely wrong to reach this function if there's no NMSettingUser set; this function **cannot** set a NMSettingUser.

pwithnall: I just reworked the code as I had in mind, and I think it’s an improvement. See what you think [here](https://github.com/endlessm/gnome-control-center/commit/8386a9170525f0326fbfae7cd2e2e9faa7537c29). Hopefully it explains my idea a bit better.

Another reason I prefer the approach I’m suggesting is that it means you don’t end up always storing the default values on the `NMSettingUser` (only if the user changes the values to something, then back to the defaults). It seems a bit wasteful to create an `NMSettingUser`, then immediately set the default values on it, given that you were already handling the lack of a return value from the `nm_setting_user_get_value()` calls.

PR #121, opened 2018-03-09
sharing: fix case in content sharing description

wjt:
This is a sentence, so it should use sentence case, where only proper nouns like “Android” are capitalized mid-sentence.

I accept that other descriptions throughout the control center are not consistent on this point. Some, but not all, follow the convention where the name of the feature is in Title Case. Regardless, “Devices” and “Network” should certainly not be capitalized.

https://phabricator.endlessm.com/T20381
https://phabricator.endlessm.com/T20668


erikos:
Thanks Will, that's great!

PR #125, opened 2018-04-13
T22075 Fix use-after-free when saving tariff information

pwithnall:
https://phabricator.endlessm.com/T22075

Compile tested only.


panels/updates/cc-updates-panel.c
```diff
@@ -156,6 +157,12 @@ store_automatic_updates_setting (NMConnection *connection,
   setting_user = ensure_setting_user (connection);
   g_assert (setting_user != NULL);
 
+  /* Calling nm_setting_user_set_data() causes notifications from NM. */
+  if (self->current_device && self->changed_id)
+    g_signal_handler_block (self->current_device, self->changed_id);
+  if (self->current_connection && self->connection_changed_id)
+    g_signal_handler_block (self->current_connection, self->connection_changed_id);
+
```
GeorgesStavracas: Do we ever call this function before connecting to these signals? It wouldn't be how I imagined its usage to be, and I'd rather have a `g_assert(self->current_device && self->changed_id > 0)` than these `if`s.

panels/updates/cc-updates-panel.c
```diff
@@ -188,17 +191,23 @@ store_automatic_updates_setting (NMConnection *connection,
...
-  g_application_hold (g_application_get_default ());
 
-  nm_remote_connection_commit_changes_async (NM_REMOTE_CONNECTION (connection),
-                                             TRUE, /* save to disk */
-                                             NULL,
-                                             on_network_changes_committed_cb,
-                                             NULL);
+  /* Only commit the changes if there were no errors. */
+  if (!errored)
+    {
```
GeorgesStavracas: This is just a matter of style, but I prefer early returns.

pwithnall: I did this to avoid having the `g_signal_handler_unblock()` calls in every early return (or in a `goto error` block).

GeorgesStavracas: What I'm saying is that I'd prefer a
```c
if (errored)
  return;
```

here

panels/updates/cc-updates-panel.c
GeorgesStavracas: I guess that's fine for now.

panels/updates/cc-updates-panel.c
pwithnall: Oh, I see. I can do that.

pwithnall:
Pushed an update to fix your comment. Waiting for Jenkins to succeed before pushing.

PR #128, opened 2018-04-24
T22241

GeorgesStavracas:
This commit does two big things:

 * Change the saving code to always save 'allow-downloads-when-metered'
   as '1', and instead save 'allow-downloads' with the value from the
   Automatic Updates switcher.

 * Add an upgrade code for users that have the old setting, but not the
   new one.

https://phabricator.endlessm.com/T22241

### DIFF
--- a/panels/updates/cc-updates-panel.c
+++ b/panels/updates/cc-updates-panel.c
@@ -27,6 +27,7 @@
 #include <NetworkManager.h>
 
 #define NM_SETTING_ALLOW_DOWNLOADS_WHEN_METERED "connection.allow-downloads-when-metered"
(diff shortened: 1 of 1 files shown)


panels/updates/cc-updates-panel.c
```diff
@@ -177,6 +179,21 @@ store_automatic_updates_setting (CcUpdatesPanel *self,
...
+                            &error);
+
+  if (error)
+    {
+      g_warning ("Error storing "NM_SETTING_ALLOW_DOWNLOADS": %s", error->message);
+      g_clear_error (&error);
+      errored = TRUE;
+    }
+
+
```
pwithnall: Nitpick: Spurious double empty line here.

PR #129, opened 2018-05-02
T22384

GeorgesStavracas:
Mogwai treats the timezones as separate entities from the date,
for the dates are treated as UNIX offsets from 0. As such, they
are always treated as UTC times.

The current tariff editor code was converting the UNIX offset
to the local time, and that was causing the timezone offset to
be applied twice.

To fix that, instead of converting the tariff timezone from
UTC to local, copy its values to a new datetime with the local
timezone.

https://phabricator.endlessm.com/T22384

### DIFF
--- a/panels/updates/cc-tariff-editor.c
+++ b/panels/updates/cc-tariff-editor.c
@@ -23,7 +23,6 @@
 
 #define CLOCK_SCHEMA     "org.gnome.desktop.interface"
 #define CLOCK_FORMAT_KEY "clock-format"
-#define SECONDS_PER_DAY  (24 * 60 * 60)
 
 struct _CcTariffEditor
 {
@@ -46,6 +45,7 @@ struct _CcTariffEditor
   guint8              time_period_from;
   guint8              time_period_to;
 
+  /* Always in UTC */
   guint64             seconds_to_start;
   guint64             seconds_to_end;
 };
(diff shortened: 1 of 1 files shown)


panels/updates/cc-tariff-editor.c
```diff
@@ -257,8 +256,18 @@ update_tariff (CcTariffEditor *self,
   /* 2. Allowed downloads period */
   allowed_start_utc = g_date_time_new_from_unix_utc (self->seconds_to_start);
   allowed_end_utc = g_date_time_new_from_unix_utc (self->seconds_to_end);
```
pwithnall: Seems to me that the problem is that `seconds_to_start` and `seconds_to_end` are in the local timezone (because they’re directly calculated from the hours/minutes the user inputs in the UI), but they’re being converted to a `GDateTime` using `g_date_time_new_from_unix_utc()`. Would it work to do
```
allowed_start = g_date_time_new_from_unix_local (self->seconds_to_start);
allowed_end = g_date_time_new_from_unix_local (self->seconds_to_end);
```

I was wondering if that opens this code up to potential failure (`g_date_time_new_from_unix_local()` returning `NULL`) if, for example, the user has set downloads to start at midnight, but they’re in a positive timezone (UTC+x) — but I don’t think it can fail, since `GDateTime` supports UTC instants down to the year 1AD, whereas a positive timezone on a UNIX timestamp of 0 would result in a time around December 1969.

It would still be useful to add `g_assert (allowed_start != NULL)` (etc.) just to make it clearer that the error handling here has been thought about.

panels/updates/cc-tariff-editor.c
```diff
@@ -89,28 +90,48 @@ get_am_hour (GDateTime *dt)
...
   guint factor_from = 0;
   guint factor_to = 0;
 
+  unix_start = g_date_time_new_from_unix_local (0);
+
   if (self->clock_format == G_DESKTOP_CLOCK_FORMAT_12H)
     {
-      factor_from = self->time_period_from == AM ? 0 : SECONDS_PER_DAY / 2;
-      factor_to = self->time_period_to == AM ? 0 : SECONDS_PER_DAY / 2;
+      factor_from = self->time_period_from == AM ? 0 : 12;
```
pwithnall: `SECONDS_PER_DAY` is no longer used, so its definition can be deleted.

panels/updates/cc-tariff-editor.c
```diff
@@ -89,28 +90,48 @@ get_am_hour (GDateTime *dt)
 static void
 update_seconds_from_adjustments (CcTariffEditor *self)
 {
+  g_autoptr(GDateTime) local_start = NULL;
+  g_autoptr(GDateTime) local_end = NULL;
+  g_autoptr(GDateTime) unix_start = NULL;
   guint factor_from = 0;
   guint factor_to = 0;
 
+  unix_start = g_date_time_new_from_unix_local (0);
```
pwithnall: This is unused.

PR #131, opened 2018-05-08
Fix bugs on clock format handling code

GeorgesStavracas:
There were 2 problems found on the clock format code:

  1. It was mistakenly putting the start hour in the end hour spinner.
  2. It was using UTC datetimes when it should've been using the local timezone.

https://phabricator.endlessm.com/T22384

### DIFF
--- a/panels/updates/cc-tariff-editor.c
+++ b/panels/updates/cc-tariff-editor.c
@@ -214,7 +214,7 @@ setup_tariff (CcTariffEditor  *self,
       gtk_stack_set_visible_child_name (GTK_STACK (self->stack_to), self->time_period_to == AM ? "am" : "pm");
 
       gtk_adjustment_set_value (self->adjustment_from_hours, get_am_hour (period_start));
-      gtk_adjustment_set_value (self->adjustment_to_hours, get_am_hour (period_start));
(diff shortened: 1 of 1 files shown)


panels/updates/cc-tariff-editor.c
```diff
@@ -301,8 +301,8 @@ update_adjustments (CcTariffEditor *self)
 
   g_debug ("Updating adjustments");
 
-  start = g_date_time_new_from_unix_utc (self->seconds_to_start);
-  end = g_date_time_new_from_unix_utc (self->seconds_to_end);
+  start = g_date_time_new_from_unix_local (self->seconds_to_start);
+  end = g_date_time_new_from_unix_local (self->seconds_to_end);
```
joaquimrocha: AFAIU, this is not related to 12h systems only but affects both formats, right? So it should be done in a separate commit to avoid confusion.

GeorgesStavracas: No, it is related to 12h. And I think it shouldn't be split in 2 commits, because it'll be squashed in the next rebase and nothing from it will be preserved, not even the commit message.

joaquimrocha: Just FTR, as discussed on Slack, it **does** affect both formats. But doing it as a fixup commit makes this less important.

PR #138, opened 2018-11-02
user-accounts: use vendor file for alternative faces directory

cosimoc:
Instead of a custom build option.

https://phabricator.endlessm.com/T23766


cosimoc:
Sorry for the back and forth, @GeorgesStavracas. You should be able to squash this with the previous commit upon rebase.

PR #144, opened 2018-12-18
background: Add support for vendor file wallpaper directory

dnohales:
https://phabricator.endlessm.com/T24867

### DIFF
--- a/panels/background/cc-background-xml.c
+++ b/panels/background/cc-background-xml.c
@@ -206,9 +206,11 @@ cc_background_xml_load_xml_internal (CcBackgroundXml *xml,
 	      bg_uri = NULL;
 	    } else {
 	      GFile *file;
-	      file = g_file_new_for_commandline_arg (content);
+	      gchar *dirname = g_path_get_dirname (filename);
+	      file = g_file_new_for_commandline_arg_and_cwd (content, dirname);
 	      bg_uri = g_file_get_uri (file);
 	      g_object_unref (file);
+	      g_free (dirname);
 	    }
 	    SET_FLAG(CC_BACKGROUND_ITEM_HAS_URI);
 	    g_object_set (G_OBJECT (item), "uri", bg_uri, NULL);
--- a/panels/background/cc-background-xml.c
+++ b/panels/background/cc-background-xml.c
@@ -19,6 +19,8 @@
  * along with this program; if not, see <http://www.gnu.org/licenses/>.
  */
 
+#include "config.h"
+
 #include <gio/gio.h>
 #include <string.h>
 #include <libxml/parser.h>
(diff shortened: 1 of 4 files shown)


panels/background/cc-background-xml.c
```diff
@@ -426,6 +429,33 @@ cc_background_xml_load_from_dir (const gchar      *path,
   g_object_unref (enumerator);
 }
 
+static gchar *
+read_vendor_background_directory (void)
+{
+  g_autoptr(GKeyFile) keyfile = NULL;
+  g_autoptr(GError) error = NULL;
+  gchar *result = NULL;
```
GeorgesStavracas: Nitpick: `g_autofree gchar *result = NULL` then `return g_steal_pointer (&result);`

panels/background/Makefile.am
```diff
@@ -11,6 +11,7 @@ AM_CPPFLAGS =						\
 	$(PANEL_CFLAGS)					\
 	$(BACKGROUND_PANEL_CFLAGS)			\
 	-DGNOMELOCALEDIR="\"$(datadir)/locale\""	\
+	-DVENDOR_CONF_FILE=\""$(VENDOR_CONF_FILE)"\"	\
```
GeorgesStavracas: Since this is apparently being shared by various panels, I guess it's about time to put this define under `PANEL_CFLAGS` (in a separate commit).

panels/background/cc-background-xml.c
dnohales: Hmm, interesting, is there any benefit/difference of doing that? (just curious)

dnohales:
Updated.

panels/background/cc-background-xml.c
```diff
@@ -434,6 +466,10 @@ cc_background_xml_load_list (CcBackgroundXml *data,
   gchar * datadir;
   gint i;
 
+  datadir = read_vendor_background_directory ();
```
cosimoc: Since this can return `NULL`, and `cc_background_xml_load_from_dir()` does not cope with that, I think you should add a `datadir != NULL` check here.

dnohales:
Updated.

panels/background/cc-background-xml.c
```diff
@@ -426,6 +431,33 @@ cc_background_xml_load_from_dir (const gchar      *path,
...
+
+    return NULL;
+  }
+
+  result = g_key_file_get_string (keyfile,
+                                  VENDOR_BACKGROUND_GROUP,
+                                  VENDOR_BACKGROUND_DIRECTORY_KEY,
+                                  &error);
+  if (!result) {
+    g_warning ("Could not read backgrounds path from %s: %s", VENDOR_CONF_FILE, error->message);
```
cosimoc: On second thought, I think we should avoid printing this warning when the group or key are not found. Something like what we did here: https://github.com/endlessm/gnome-control-center/blob/5fe3074928bfe07fe33790d900386c2d6529c6aa/panels/user-accounts/um-photo-dialog.c#L410

dnohales: We should move to a common function eventually.

dnohales:
Updated

PR #147, opened 2019-01-04
eos3.5: Add App Permissions to User Accounts panel

wjt:
This is the patches from #141, cherry-picked for eos3.5 to test CI while master CI is not in its happy place.


wjt:
test this please

PR #151, opened 2019-01-08
T23897 Mostly fix age mappings with parental controls

pwithnall:
This is based on #150. Only the top two commits are relevant here.

We store parental controls age restrictions as a set of OARS categories, but the user interface in g-c-c presents a locale-specific set of ratings to the user. So we need a reversible mapping from locale-specific rating system → CSM → OARS → CSM → locale-specific rating system.

Unfortunately, mostly due to the fact that CSM doesn’t have OARS ↔ CSM mappings for age 16, that’s not possible for all ratings systems. In particular, IARC and PEGI had problems.

Fix that by adding some workarounds to the reverse mapping.

Also update the ESRA categories while we’re there, since they seem to have changed in the last 3 years. This fixes an issue with round-tripping their old ‘+25’ category (it no longer exists).

There are still issues with CERO rating D (gets mapped back to CERO C because there is no CSM age 17) and with OFLCNZ R16 (gets mapped back to R15 because there is no CSM age 16). These are not critical for the regions we care about most immediately. The long-term fix is to ensure that there are OARS ↔ CSM mappings for ages 16 and 17, but that requires discussion on the specification.

https://phabricator.endlessm.com/T23897

### DIFF
--- a/panels/user-accounts/gs-content-rating.c
+++ b/panels/user-accounts/gs-content-rating.c
@@ -22,6 +22,7 @@
 #include "config.h"
 
 #include <glib/gi18n.h>
+#include <string.h>
 
 #include "gs-content-rating.h"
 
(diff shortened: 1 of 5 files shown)


pwithnall:
https://github.com/endlessm/gnome-control-center/pull/152 is available as an alternative (mutually exclusive) approach to solving this problem.

PR #152, opened 2019-01-08
T23897 Different approach to fixing age mapping with parental controls

pwithnall:
This is an alternative to https://github.com/endlessm/gnome-control-center/pull/151 (mutually exclusive).

This approach adds some additional CSM ↔ OARS mappings so that mappings exist for CSM ages 16 and 17, which are mapped to from several locale-specific ratings systems.

It fixes all mappings (unlike #151, which leaves one OFLCNZ R16 and CERO D broken), but at the cost of diverging from upstream appstream-glib’s set of CSM ↔ OARS mappings.

This is based on https://github.com/endlessm/gnome-control-center/pull/150, so only the top two commits are relevant.

https://phabricator.endlessm.com/T23897

### DIFF
--- a/panels/user-accounts/gs-content-rating.c
+++ b/panels/user-accounts/gs-content-rating.c
@@ -22,6 +22,7 @@
 #include "config.h"
 
 #include <glib/gi18n.h>
+#include <string.h>
 
 #include "gs-content-rating.h"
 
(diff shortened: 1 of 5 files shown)


GeorgesStavracas:
I prefer this approach as well.

PR #155, opened 2019-01-17
T25063 Parental controls usability improvements

pwithnall:
Various minor usability improvements from https://phabricator.endlessm.com/T25063.

Includes various translatable string changes,.

### DIFF
--- a/panels/user-accounts/data/app-permissions.ui
+++ b/panels/user-accounts/data/app-permissions.ui
@@ -27,7 +27,7 @@
       <object class="GtkLabel">
         <property name="visible">True</property>
         <property name="xalign">0.0</property>
-        <property name="label" translatable="yes">Prevent this user from opening some apps.</property>
+        <property name="label" translatable="yes">Prevent this user from opening some apps by turning them off below.</property>
         <attributes>
           <attribute name="scale" value="0.83333" />
         </attributes>
--- a/panels/user-accounts/data/app-permissions.ui
+++ b/panels/user-accounts/data/app-permissions.ui
@@ -28,6 +28,8 @@
         <property name="visible">True</property>
         <property name="xalign">0.0</property>
         <property name="label" translatable="yes">Prevent this user from opening some apps by turning them off below.</property>
+        <property name="use-underline">True</property>
+        <property name="mnemonic-widget">listbox</property>
         <attributes>
           <attribute name="scale" value="0.83333" />
         </attributes>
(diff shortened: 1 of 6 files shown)


pwithnall:
Submitted the carousel changes upstream here: https://gitlab.gnome.org/GNOME/gnome-control-center/merge_requests/352

panels/user-accounts/data/user-accounts-dialog.ui
```diff
@@ -106,329 +106,336 @@
               </packing>
             </child>
             <child>
-              <object class="GtkGrid" id="user-options">
+              <object class="GtkScrolledWindow">
```
wjt: Do you also want to remove the scrolled window from app-permissions.ui? It might feel a little weird to have nested scrollable areas?

panels/user-accounts/data/app-permissions.ui
```diff
@@ -126,7 +130,9 @@
           <object class="GtkLabel" id="system_installation_label">
             <property name="visible">True</property>
             <property name="xalign">1.0</property>
-            <property name="label" translatable="yes">Installed apps can be used by every user:</property>
+            <property name="label" translatable="yes">Install Apps for All _Users:</property>
```
wjt: The argument for changing this string is basically that, when translated into at least French and Portuguese, it forces the window to expand horizontally:

screenshot from 2019-01-18 10-23-56
captura de tela de 2019-01-18 10-45-09

wjt:
@GeorgesStavracas when would you prefer they be changed? Only for 3.6? I guess I agree that none of the problems here are so bad that we couldn't live with them till then…

panels/user-accounts/data/user-accounts-dialog.ui
pwithnall: The one for ‘Restrict Apps’? No, I think that would make the dialogue very long and look quite weird.

panels/user-accounts/data/app-permissions.ui
pwithnall: The other argument for changing it is that it isn’t in Title Case and can’t be mnemonic-ised properly in its previous state.

pwithnall:
> not really keen on changing translatable strings now

Why not?

> I guess I agree that none of the problems here are so bad that we couldn't live with them till then…

Or we could change the strings in English and live with the problems in other languages, if we don’t want to get another round of translations done.

PR #159, opened 2019-01-22
user-accounts: Allow ‘Remove User’ button to be focused

pwithnall:
Otherwise it isn’t included in the tab focus cycle, and can’t be
navigated to with the tab key.

https://phabricator.endlessm.com/T24990


pwithnall:
Submitted upstream here too: https://gitlab.gnome.org/GNOME/gnome-control-center/merge_requests/357

GeorgesStavracas:
Merged upstream, and here too

PR #158, opened 2019-01-22
T25171 Reload parental controls apps list when apps are installed/uninstalled

pwithnall:
This is based on the branch for #157. Only the top commit is relevant.

https://phabricator.endlessm.com/T25171

PR #160, opened 2019-01-22
T25172 user-accounts: Hide apps with duplicate flatpak IDs

pwithnall:
App blacklisting is done on the basis of the flatpak ID of an app.
LibreOffice has a single flatpak ID shared between all its components,
but installs multiple .desktop files with different desktop file IDs, so
they get listed separately in the parental controls app list. However,
blacklisting one of them toggles the blacklist status for all other
LibreOffice components, which is not an expected user behaviour.

Fix this by hiding all apps which have the same flatpak ID as one
already listed. Select the one to show as the one with the shortest
desktop file ID, since this will typically be the ‘main’ .desktop file
whose ID is a prefix of all the others.

https://phabricator.endlessm.com/T25172

---

Only the top-most commit is relevant here. This branch is based on #158 to avoid merge conflicts.

### DIFF
--- a/panels/user-accounts/um-app-permissions.c
+++ b/panels/user-accounts/um-app-permissions.c
@@ -136,17 +136,50 @@ static const gchar * const oars_categories[] =
 
 /* Auxiliary methods */
 
+static gint
+app_compare_id_length_cb (gconstpointer a,
+                          gconstpointer b)
+{
+  GAppInfo *info_a = (GAppInfo *) a, *info_b = (GAppInfo *) b;
+  const gchar *id_a, *id_b;
+
+  id_a = g_app_info_get_id (info_a);
+  id_b = g_app_info_get_id (info_b);
+
+  if (id_a == NULL && id_b == NULL)
+    return 0;
+  else if (id_a == NULL)
+    return -1;
+  else if (id_b == NULL)
+    return 1;
+
+  return strlen (id_a) - strlen (id_b);
+}
+
 static void
 reload_apps (UmAppPermissions *self)
 {
   GList *iter, *apps;
+  g_autoptr(GHashTable) seen_flatpak_ids = NULL;
 
   apps = g_app_info_get_all ();
+
+  /* Sort the apps by increasing length of #GAppInfo ID. When coupled with the
+   * deduplication of flatpak IDs, below, this should ensure that we pick the
+   * ‘base’ app out of any set with matching prefixes and identical app IDs, and
+   * show only that.
+   *
+   * This is designed to avoid listing all the components of LibreOffice, which
+   * all share an app ID and hence have the same entry in the parental controls
+   * app filter. */
+  apps = g_list_sort (apps, app_compare_id_length_cb);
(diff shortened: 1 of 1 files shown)


panels/user-accounts/um-app-permissions.c
```diff
@@ -134,19 +134,52 @@ static const gchar * const oars_categories[] =
...
+
+  id_a = g_app_info_get_id (info_a);
+  id_b = g_app_info_get_id (info_b);
+
+  if (id_a == NULL && id_b == NULL)
+    return 0;
+  else if (id_a == NULL)
+    return -1;
+  else if (id_b == NULL)
+    return 1;
```
wjt: It's a shame one can't use:

```c
if (id_a == NULL || id_b == NULL)
  return id_a - id_b;
```

because pointer arithmetic returns a type wider than `int`... :-)

pwithnall:
Needs rebasing once #158 is merged, but still good to review.

panels/user-accounts/um-app-permissions.c
pwithnall: I think that’s not particularly clear/obvious to read.

wjt: Not a serious suggestion!

panels/user-accounts/um-app-permissions.c
```diff
@@ -134,19 +134,52 @@ static const gchar * const oars_categories[] =
...
+  apps = g_list_sort (apps, app_compare_id_length_cb);
+  seen_flatpak_ids = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
+
   g_list_store_remove_all (self->apps);
 
   for (iter = apps; iter; iter = iter->next)
     {
       GAppInfo *app;
       const gchar *app_name;
+      g_autofree gchar *flatpak_id = NULL;
```
GeorgesStavracas: Nitpick: `g_autofree` and `g_autoprt` variables generally are at the top of the variable declaration section.

panels/user-accounts/um-app-permissions.c
```diff
@@ -164,7 +197,18 @@ reload_apps (UmAppPermissions *self)
           continue;
         }
 
-      g_debug ("Processing app '%s' (%s)", g_app_info_get_id (app), g_app_info_get_executable (app));
+      flatpak_id = g_desktop_app_info_get_string (G_DESKTOP_APP_INFO (app), "X-Flatpak");
+      g_debug ("Processing app ‘%s’ (Exec=%s, X-Flatpak=%s)",
+               g_app_info_get_id (app), g_app_info_get_executable (app),
+               flatpak_id);
```
GeorgesStavracas: Nitpick: one argument per line here.

panels/user-accounts/um-app-permissions.c
pwithnall: OOI, why?

pwithnall:
Updated.

pwithnall:
The earlier branches have been merged. Rebased this one, and merging it as per the CR+ earlier.

PR #163, opened 2019-01-25
search: Filter list of search providers according to parental controls

pwithnall:
If a non-admin user is using the ‘Search’ panel in the control center,
it doesn’t make sense to show them preferences for apps which are
blacklisted for them.

https://phabricator.endlessm.com/T25079


panels/search/cc-search-panel.c
```diff
@@ -657,10 +696,27 @@ search_providers_discover_thread (GTask *task,
                                   gpointer task_data,
                                   GCancellable *cancellable)
 {
+  DiscoverProvidersResult *result = NULL;
   GList *providers = NULL;
   const gchar * const *system_data_dirs;
   int idx;
+  EpcAppFilter *filter = NULL;
```
GeorgesStavracas: Is it not possible to use `g_autoptr(EpcAppFilter)`?

panels/search/cc-search-panel.c
```diff
@@ -657,10 +696,27 @@ search_providers_discover_thread (GTask *task,
...
+  GError *local_error = NULL;
+
+  /* Load the user’s parental controls settings too, so we can filter the list. */
+  filter = epc_get_app_filter (NULL,
+                               getuid (),
+                               FALSE,
+                               cancellable,
+                               &local_error);
```
GeorgesStavracas: Perhaps it's better to create the app filter at panel initialization time.

Notice that at the next Settings rebase, a new mechanism for storing and sharing objects across panels is in place, and we'll be able to use it to only load the app filter once.

panels/search/cc-search-panel.c
pwithnall: None of the other code in `cc-search-panel.c` uses it, so I was being consistent. But I can use `g_autoptr()` there if you would prefer.

pwithnall:
I’ve updated the branch to use `g_autoptr()`. Note that **this uses a fixup commit** so don’t push if you ack it.

PR #165, opened 2019-02-13
T25360

GeorgesStavracas:
https://phabricator.endlessm.com/T25360

### DIFF
--- a/panels/user-accounts/data/account-dialog.ui
+++ b/panels/user-accounts/data/account-dialog.ui
@@ -155,6 +155,7 @@
                 <child>
                   <object class="GtkEntry" id="local_name">
                     <property name="visible">True</property>
+                    <property name="max-length">255</property>
                     <property name="can_focus">True</property>
                     <property name="activates_default">True</property>
                     <property name="hexpand">True</property>
--- a/panels/user-accounts/data/account-dialog.ui
+++ b/panels/user-accounts/data/account-dialog.ui
@@ -155,7 +155,7 @@
                 <child>
                   <object class="GtkEntry" id="local_name">
                     <property name="visible">True</property>
-                    <property name="max-length">255</property>
+                    <property name="max-length">80</property>
                     <property name="can_focus">True</property>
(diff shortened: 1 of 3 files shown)


panels/user-accounts/data/user-accounts-dialog.ui
```diff
@@ -187,6 +187,7 @@
                     <child>
                       <object class="GtkEntry" id="full-name-entry">
                         <property name="visible">True</property>
+                        <property name="max-length">255</property>
```
wjt: I think this one is the one used when creating an account. I did a little testing and it appears that accountsservice's behaviour on Debian-like systems is different between creating an account and changing an existing account's name:

* When creating an account, the name cannot be longer than 80 bytes (not characters: 80 'a's was accepted, 79 'a's followed by an 'á', which is 81 bytes of UTF-8, was not)
* When modifying an existing account's name, anything goes

At the very least we should change max-length to 80 for the account creation case here (matching https://github.com/endlessm/gnome-initial-setup/commit/84d165f80c37da208cce76d00c4d974ec07c02e9 ); we should open a follow-up task to either try to make accountsservice accept >80 bytes when creating accounts, or implement the limit as bytes rather than chars in both g-i-s and g-c-c.

GeorgesStavracas: Set to 80 characters.

panels/user-accounts/data/account-dialog.ui
```diff
@@ -155,6 +155,7 @@
                 <child>
                   <object class="GtkEntry" id="local_name">
                     <property name="visible">True</property>
+                    <property name="max-length">80</property>
```
wjt: I think these would be better changed as a separate patch. When you rebase this module in a few weeks you'll have to remember to apply this change as a separate patch anyway, no?

GeorgesStavracas: Updated.

PR #166, opened 2019-02-14
background: Stop using a custom XDG dir for wallpapers

pwithnall:
This is a partial revert of commit
84eca50c748fe982392bf9bad8084107061f8918. It drops
gnome-wallpapers.desktop as a way of specifying where g-c-c should look
for wallpapers, and instead hard-codes the directory to look in as a
subdirectory of XDG_PICTURES_DIR.

There are several reasons for this:
 • It was using a custom GLib API
   (g_get_user_special_dir_for_desktop_id()) which was not upstream and
   wasn’t going to be accepted upstreamed.
 • Looking at `man 5 user-dirs.dirs`, it says “the format of
   user-dirs.dirs is designed to allow direct sourcing of this file in
   shell scripts”, but if gnome-wallpapers.desktop were ever overridden
   in user-dirs.dirs, that would break, as `gnome-wallpapers.desktop` is
   not a valid sh variable name. So the default value could only ever be
   used.
 • Nautilus’ ‘Set as Wallpaper’ feature, which is what this was designed
   to complement, always copies images to
   ${XDG_PICTURES_DIR}/Wallpapers rather than using
   gnome-wallpapers.desktop. (The upstream version of Nautilus
   translates the ‘Wallpapers’ directory; the EOS 3.5 version does not.
   That will change.)

https://phabricator.endlessm.com/T25445

### DIFF
--- a/panels/background/Makefile.am
+++ b/panels/background/Makefile.am
@@ -94,11 +94,7 @@ desktopdir = $(datadir)/applications
 desktop_in_files = gnome-background-panel.desktop.in
 desktop_DATA = $(desktop_in_files:.desktop.in=.desktop)
 
-xdgdirsdir = $(datadir)/xdg-user-dirs
-xdgdirs_in_files = gnome-wallpapers.desktop.in
-xdgdirs_DATA = $(xdgdirs_in_files:.desktop.in=.desktop)
-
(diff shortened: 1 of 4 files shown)


GeorgesStavracas:
retest this please

pwithnall:
Associated `debian-master` change: https://github.com/endlessm/gnome-control-center/pull/167

PR #174, opened 2019-06-07
Allow filtering apps with X-Parental-Controls desktop file entry

andrunko:
https://phabricator.endlessm.com/T26957

### DIFF
--- a/panels/user-accounts/cc-app-permissions.c
+++ b/panels/user-accounts/cc-app-permissions.c
@@ -161,25 +161,26 @@ reload_apps (CcAppPermissions *self)
 {
   GList *iter, *apps;
   g_autoptr(GHashTable) seen_flatpak_ids = NULL;
+  g_autoptr(GHashTable) seen_executables = NULL;
 
   apps = g_app_info_get_all ();
 
   /* Sort the apps by increasing length of #GAppInfo ID. When coupled with the
-   * deduplication of flatpak IDs, below, this should ensure that we pick the
-   * ‘base’ app out of any set with matching prefixes and identical app IDs, and
-   * show only that.
+   * deduplication of flatpak IDs and executable paths, below, this should ensure that we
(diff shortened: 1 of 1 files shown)


panels/search/cc-search-panel.c
```diff
@@ -614,7 +614,7 @@ search_providers_discover_ready (GObject *source,
 
   for (l = discover_result->providers; l != NULL; l = l->next)
     {
-      g_autoptr(GFile) provider = l->data;
+      GFile *provider = l->data;
```
pwithnall: This seems to have been a regression in the rebase of commit c773723d78, and should be squashed into that next time we rebase. I’ve checked, and the bug doesn’t seem to be present on the `eos3.6` branch.

panels/user-accounts/cc-app-permissions.c
```diff
@@ -191,26 +192,54 @@ reload_apps (CcAppPermissions *self)
...
+        }
+      else if (g_desktop_app_info_has_key (G_DESKTOP_APP_INFO (app), "X-Parental-Controls"))
         {
-          g_debug (" → Skipping ‘%s’ due to seeing its flatpak ID already",
-                   g_app_info_get_id (app));
-          continue;
+          g_autofree gchar *parental_controls_type = NULL;
+          g_autofree gchar *executable = NULL;
+
+          parental_controls_type = g_desktop_app_info_get_string (G_DESKTOP_APP_INFO (app),
```
pwithnall: For future-proofness, maybe ignore desktop files which have `X-Parental-Controls=none`?

panels/search/cc-search-panel.c
pwithnall: (Now proposed in a separate PR as https://github.com/endlessm/gnome-control-center/pull/175.)

panels/user-accounts/cc-app-permissions.c
andrunko: Done.

PR #176, opened 2019-06-11
parental-controls: Migrate to malcontent

uajain:
eos-parental-controls is renamed to malcontent for upstreaming
purposes hence, we should move to libmalcontent so that it can
be dropped from Endless package-management system as well.

This commit does not introduce any functional changes. In the next
rebase cycle, this commit can be squashed with the commit where the
parental controls is introduced (c315ab1ed).

### DIFF
--- a/panels/search/cc-search-panel.c
+++ b/panels/search/cc-search-panel.c
@@ -25,7 +25,7 @@
 
 #include <gio/gdesktopappinfo.h>
 #include <glib/gi18n.h>
-#include <libeos-parental-controls/app-filter.h>
+#include <libmalcontent/malcontent.h>
 
 struct _CcSearchPanel
 {
@@ -44,7 +44,7 @@ struct _CcSearchPanel
 
   CcSearchLocationsDialog  *locations_dialog;
 
-  EpcAppFilter *filter;  /* (owned) (nullable) */
+  MctAppFilter *filter;  /* (owned) (nullable) */
 };
 
 CC_PANEL_REGISTER (CcSearchPanel, cc_search_panel)
@@ -548,7 +548,7 @@ search_panel_add_one_provider (CcSearchPanel *self,
    * the current user’s parental controls settings. */
   g_assert (self->filter != NULL);
 
-  if (!epc_app_filter_is_appinfo_allowed (self->filter, app_info))
+  if (!mct_app_filter_is_appinfo_allowed (self->filter, app_info))
     {
       g_debug ("Filtering out search provider from application ‘%s’ as it’s "
                "blacklisted by parental controls.", g_app_info_get_id (app_info));
@@ -565,13 +565,13 @@ search_panel_add_one_provider (CcSearchPanel *self,
 typedef struct
 {
   GList *providers;  /* (element-type GFile) (owned) */
-  EpcAppFilter *filter;  /* (owned) */
+  MctAppFilter *filter;  /* (owned) */
 } DiscoverProvidersResult;
 
 static void
 discover_providers_result_free (DiscoverProvidersResult *result)
 {
-  g_clear_pointer (&result->filter, epc_app_filter_unref);
+  g_clear_pointer (&result->filter, mct_app_filter_unref);
   g_list_free_full (result->providers, g_object_unref);
   g_free (result);
 }
@@ -603,7 +603,7 @@ search_providers_discover_ready (GObject *source,
 
   /* Store the app filter. */
   g_assert (self->filter == NULL);
-  self->filter = epc_app_filter_ref (discover_result->filter);
+  self->filter = mct_app_filter_ref (discover_result->filter);
 
   /* Process the providers. */
   if (discover_result->providers == NULL)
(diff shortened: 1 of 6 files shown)


uajain:
This depends on @andrunko's [eos-parental-control pull-request](https://github.com/endlessm/eos-parental-controls/pull/24/files#diff-d837a1c17de0268bcea321239ddbed47) which solves package dependency. I locally tested that PR and  then briefly tested this branch for enabling/disabling various apps for a `child` user. I didn't faced any problem.

Hence, DO NOT MERGE until  [eos-parental-control pull-request](https://github.com/endlessm/eos-parental-controls/pull/24/files#diff-d837a1c17de0268bcea321239ddbed47) is merged.

panels/search/cc-search-panel.c
```diff
@@ -677,18 +677,28 @@ search_providers_discover_thread (GTask *task,
                                   GCancellable *cancellable)
 {
   g_autoptr(DiscoverProvidersResult) result = NULL;
+  g_autoptr(GDBusConnection) system_bus = NULL;
+  MctManager *manager;
```
pwithnall: Could use `g_autoptr(MctManager) manager = NULL`, no?

panels/search/cc-search-panel.c
```diff
@@ -677,18 +677,28 @@ search_providers_discover_thread (GTask *task,
...
+  manager = mct_manager_new (system_bus);
   /* Load the user’s parental controls settings too, so we can filter the list. */
-  filter = epc_get_app_filter (NULL,
-                               getuid (),
-                               FALSE,
-                               cancellable,
-                               &local_error);
+  filter = mct_manager_get_app_filter (manager,
+                                       getuid (),
+                                       MCT_GET_APP_FILTER_FLAGS_INTERACTIVE,
```
pwithnall: The use of the `INTERACTIVE` flag here is a behaviour change from the old code. The old code passed `FALSE` to the `allow_interactive_authorization` argument of `epc_get_app_filter()`.

panels/user-accounts/cc-app-permissions.c
```diff
@@ -268,18 +269,18 @@ update_app_filter (CcAppPermissions *self)
...
 
   /* FIXME: make it asynchronous */
-  self->filter = epc_get_app_filter (NULL,
-                                     act_user_get_uid (self->user),
-                                     FALSE,
-                                     self->cancellable,
-                                     &error);
+  self->filter = mct_manager_get_app_filter (self->manager,
+                                             act_user_get_uid (self->user),
+                                             MCT_GET_APP_FILTER_FLAGS_INTERACTIVE,
```
pwithnall: Same here. I think this should not be `INTERACTIVE`.

panels/user-accounts/cc-app-permissions.c
```diff
@@ -433,7 +434,18 @@ update_allow_app_installation (CcAppPermissions *self)
...
 setup_parental_control_settings (CcAppPermissions *self)
 {
+  g_autoptr(GDBusConnection) system_bus = NULL;
   gboolean is_authorized, user_is_administrator;
+  g_autoptr(GError) error = NULL;
+
+  system_bus = g_bus_get_sync (G_BUS_TYPE_SYSTEM, self->cancellable, &error);
+  if (system_bus == NULL)
+    {
+      g_warning ("Error getting system bus for setting up parental controls: %s", error->message);
```
pwithnall: Rather than introducing the possibility of emitting a warning here, would it be possible to set up a single `GDBusConnection` instance for the system bus when the panel is first instantiated, and just use it here?

panels/user-accounts/cc-app-permissions.c
```diff
@@ -433,7 +434,18 @@ update_allow_app_installation (CcAppPermissions *self)
...
+      return;
+    }
+
+  self->manager = mct_manager_new (system_bus);
```
pwithnall: I suspect this line of code might get executed multiple times (`setup_parental_control_settings()` gets called from a few places) and hence will leak old `manager` instances. You should only need to construct the `manager` once, when the panel is first instantiated.

panels/user-accounts/cc-app-permissions.c
```diff
@@ -592,18 +604,18 @@ blacklist_apps_cb (gpointer data)
...
-  epc_set_app_filter (NULL,
-                      act_user_get_uid (self->user),
-                      new_filter,
-                      TRUE,
-                      self->cancellable,
-                      &error);
+  mct_manager_set_app_filter (self->manager,
+                              act_user_get_uid (self->user),
+                              new_filter,
+                              MCT_GET_APP_FILTER_FLAGS_INTERACTIVE,
```
pwithnall: This use of `INTERACTIVE` is correct as per the old code.

panels/user-accounts/cc-app-permissions.c
uajain: > You should only need to construct the manager once, when the panel is first instantiated.

Do you mean when "Users" panels is instantiated (in `panels/user-accounts/cc-user-panel.c`) or when the widget CcAppPermissions is instantiated (in `panels/user-accountscc-app-permissions.c`) ?

panels/user-accounts/cc-app-permissions.c
```diff
@@ -59,7 +59,8 @@ struct _CcAppPermissions
   GListStore *apps; /* (owned) */
 
   GCancellable *cancellable; /* (owned) */
-  EpcAppFilter *filter; /* (owned) */
+  MctManager   *manager; /* (owned) */
```
uajain: I forgot to cleanup this member in `_finalize`

panels/user-accounts/cc-app-permissions.c
pwithnall: Probably the latter, but whatever makes the most sense to you. I think they probably are both instantiated at roughly the same time?

uajain: Right, addressed in the latest iteration of the patch. Picked up the latter approach as that was most sensible to me. Thanks for this pointer here.

panels/user-accounts/cc-app-permissions.c
uajain: Yes, we now instantiate `MctManager` in `cc_app_permissions_init`. This entire hunk is entirely moved to `_init ()` in the latest iteration of the patch.

panels/user-accounts/cc-app-permissions.c
```diff
@@ -924,11 +926,25 @@ cc_app_permissions_class_init (CcAppPermissionsClass *klass)
...
+  g_autoptr(GError) error = NULL;
+
   gtk_widget_init_template (GTK_WIDGET (self));
 
   self->system_installation = flatpak_installation_new_system (NULL, NULL);
   self->user_installation = flatpak_installation_new_user (NULL, NULL);
 
+  self->cancellable = g_cancellable_new ();
+
+  system_bus = g_bus_get_sync (G_BUS_TYPE_SYSTEM, self->cancellable, &error);
```
pwithnall: It’s not great to do synchronous IPC in the init function of a widget, but if you add a `FIXME` comment this will probably do for now, given that a lot of other things in g-c-c don’t do async properly already.

uajain: Introduced a `FIXME`.
Thanks for the review.
Merging after jenkins-build.

PR #177, opened 2019-06-11
Drop eos-parental-controls dependency in favour of malcontent

uajain:


uajain:
Forced-pushed contains the commit message change for linking to https://phabricator.endlessm.com/T26995 instead of https://phabricator.endlessm.com/T26537

PR #178, opened 2019-06-13
applications: Filter out "eos-link-" Endless-specific webapps

uajain:
A web app can be added to desktop for easy access via:
Right-click the desktop → Add Website...
These web app have app-id prefixed with "eos-link-".

gnome-control-center 3.32 introduced a new "Applications"
panels for applications' managing details (managing permissions,
disk space usage etc.). The enormous list of "eos-link-"
webapps ended up in this applications panel where basically
have very minimal information to show/manage. Also, many of
them are non-English speaking apps, which inhibits the purpose
of search in this particular panel.

https://phabricator.endlessm.com/T27007

### DIFF
--- a/panels/applications/cc-applications-panel.c
+++ b/panels/applications/cc-applications-panel.c
@@ -39,6 +39,7 @@
 #define MASTER_SCHEMA "org.gnome.desktop.notifications"
 #define APP_SCHEMA MASTER_SCHEMA ".application"
 #define APP_PREFIX "/org/gnome/desktop/notifications/application/"
+#define EOS_LINK_PREFIX "eos-link-"
 
 struct _CcApplicationsPanel
 {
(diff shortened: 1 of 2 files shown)


panels/applications/cc-applications-panel.c
```diff
@@ -1410,13 +1411,14 @@ populate_applications (CcApplicationsPanel *self)
       GtkWidget *row;
       g_autofree gchar *id = NULL;
 
-      if (!g_app_info_should_show (info))
+      id = get_app_id (info);
+      if (!g_app_info_should_show (info) ||
+          g_str_has_prefix (id, EOS_LINK_PREFIX))
```
GeorgesStavracas: Nitpick: either keep it in a single line (preferred), or add curly braces to the if condition.

PR #179, opened 2019-06-14
applications: Enforce Parental Controls settings

uajain:
https://phabricator.endlessm.com/T27010

### DIFF
--- a/panels/applications/cc-applications-panel.c
+++ b/panels/applications/cc-applications-panel.c
@@ -22,6 +22,7 @@
 
 #include <config.h>
 #include <glib/gi18n.h>
+#include <libmalcontent/malcontent.h>
 
 #include <gio/gdesktopappinfo.h>
 
@@ -52,6 +53,10 @@ struct _CcApplicationsPanel
   gulong           monitor_id;
 
   GCancellable    *cancellable;
+
+  MctAppFilter    *app_filter;
+  MctManager      *manager;
+  guint            app_filter_id;
 
   gchar           *current_app_id;
   gchar           *current_flatpak_id;
@@ -1403,6 +1408,7 @@ populate_applications (CcApplicationsPanel *self)
 
   container_remove_all (GTK_CONTAINER (self->sidebar_listbox));
 
+  g_signal_handler_block (self->manager, self->app_filter_id);
   infos = g_app_info_get_all ();
 
   for (l = infos; l; l = l->next)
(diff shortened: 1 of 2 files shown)


panels/applications/cc-applications-panel.c
```diff
@@ -1403,6 +1408,7 @@ populate_applications (CcApplicationsPanel *self)
 
   container_remove_all (GTK_CONTAINER (self->sidebar_listbox));
 
+  g_signal_handler_block (self->manager, self->app_filter_id);
```
pwithnall: Why block this signal handler here? Do you have reason to believe that the main context might iterate while we’re inside this function, and hence end up with nested `populate_applications()` calls?

panels/applications/cc-applications-panel.c
```diff
@@ -1715,6 +1745,30 @@ cc_applications_panel_init (CcApplicationsPanel *self)
...
+  system_bus = g_bus_get_sync (G_BUS_TYPE_SYSTEM, self->cancellable, &error);
+  if (system_bus == NULL)
+    {
+      g_warning ("Error getting system bus while setting up app permissions: %s", error->message);
+      return;
+    }
+
+  /* Load the user’s parental controls settings too, so we can filter the list. */
+  self->manager = mct_manager_new (system_bus);
+  self->app_filter = mct_manager_get_app_filter (self->manager,
```
pwithnall: I don’t really like these sync calls in the constructor, but since g-c-c doesn’t set a better precedent anywhere else (that I know of), we’d end up doing a lot of work to make this async, which isn’t worth it at the moment.

So the sync wheel keeps turning, and the async one never gets started.

uajain: Yeah, I agree very much to this. `_async`-fying as downstream-first, might makes the next rebase painful to some extend because code-bits coming from upstream will get moved a lot too as a result. Hence, not looking forward to it at all during rebases :(

panels/applications/cc-applications-panel.c
uajain: Yes, that's precisely is my line of thought. If `app-filter-changed` emits at very short intervals (I don't know if this happens in real-life scenario), it might end up with long tail of `populate_applications()`.

pwithnall: OK. I don’t think that’s possible, but blocking/unblocking the signal doesn’t hurt.

PR #182, opened 2019-06-26
parental controls: Add setting to disable all browsers

andrunko:
Note that in this implementation individual browsers are not listed in the restrict apps list and can only be disabled via the `Web Browsers` switch. This was done on purpose as supporting both options to disable browsers would require more discussions on the UI/usability side of things. See https://phabricator.endlessm.com/T26895#730545 for more details.

https://phabricator.endlessm.com/T27064

### DIFF
--- a/panels/user-accounts/cc-app-permissions.c
+++ b/panels/user-accounts/cc-app-permissions.c
@@ -29,6 +29,8 @@
 
 #include "cc-app-permissions.h"
 
+#define WEB_BROWSERS_CONTENT_TYPE "x-scheme-handler/http"
+
 /* The value which we store as an age to indicate that OARS filtering is disabled. */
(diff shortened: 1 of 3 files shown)


andrunko:
> Looks good to me. Have you tested that the keyboard tab navigation works OK?
> 
Thanks, yes, I just tested and keyboard nav works fine.

> Also, I noticed from the screenshot that some of the UI labels have colons between the label and the controls (e.g. “Web Browsers:”) and some don’t. Looking at other panels in g-c-c, it looks like the colon shouldn’t be there. Fancy fixing that too? (I realise it’s not your bug, and it could be handled as a separate PR.)

Good catch, I will fix it as a separate PR.

Merging this one.

PR #184, opened 2019-06-27
parental controls: Remove colons from UI labels

andrunko:
Signed-off-by: Andre Moreira Magalhaes <andre@endlessm.com>


andrunko:
See also https://github.com/endlessm/gnome-control-center/pull/182#pullrequestreview-255091507.

andrunko:
> Are there any translations of these strings which you could update at the same time, to avoid them getting lost?

Didnt find any - I grepped for these strings in the whole source and they only appear in this `.ui` file.

PR #185, opened 2019-06-28
parental controls: Properly initialize selected age

andrunko:

### DIFF
--- a/panels/user-accounts/cc-app-permissions.c
+++ b/panels/user-accounts/cc-app-permissions.c
@@ -1009,6 +1009,7 @@ cc_app_permissions_init (CcAppPermissions *self)
 
   gtk_widget_init_template (GTK_WIDGET (self));
 
+  self->selected_age = (guint) -1;
   self->system_installation = flatpak_installation_new_system (NULL, NULL);
   self->user_installation = flatpak_installation_new_user (NULL, NULL);
 


panels/user-accounts/cc-app-permissions.c
```diff
@@ -32,7 +32,7 @@
 #define WEB_BROWSERS_CONTENT_TYPE "x-scheme-handler/http"
 
 /* The value which we store as an age to indicate that OARS filtering is disabled. */
-static const guint32 oars_disabled_age = (guint32) -1;
+static const guint oars_disabled_age = (guint) -1;
```
uajain: Drive-by comment:

What is the premise of assigning a -ve value to `guint` ?

andrunko: It was already there and I didnt want to change all the code to use gint instead (and age can't be negative anyway). So -1 represents an "unset" value. The age is compared also with values from gs_utils_content_rating_get_ages which return guint.

pwithnall: I don’t think it’s correct to change this from a `guint32` to a `guint`. The age passes through the `GAction`s as a `uint32` variant, so the `disabled_age` there is `(guint32) -1`, which differs from `(guint) -1` on 64-bit systems. Grep for `set-age` in this C file.

andrunko: hmm, yes, I noticed that too but I think it will still do the right thing as it prints the value using `%u` to a string - at least on a simple test code printing `(guint) -1` and `(guint32) -1` as `%u` gives me the same value.

The thing is this code was already mixing guint32/guint in different places, hence why I changed, for example, on `on_set_age_action_activated()` we have `age` declared as `guint` but the value is set to `g_variant_get_uint32()` and then compared to the disabled age (which with the change is also `guint`). The `oars_disabled_value` is also compared to `selected_age` which was always a `guint`, so not sure here, although I think it will do the right thing.

andrunko: Another thing is that the code setting the `GAction` (`set-age`) was already setting it from `guint` when setting the ages returned by `gs_utils_content_rating_get_ages()` (which returns a `guint` list).

pwithnall: > `(guint32) -1`, which differs from `(guint) -1` on 64-bit systems

I was very wrong to say this. `guint` is 32 bits on 64-bit systems. `guint64` obviously isn’t. So `guint32` and `guint` should be equivalent on 32-bit and 64-bit systems.

However, I still think it’s wrong to change the type of `oars_disabled_age`. It’s not needed as part of the fix (as the types are equivalent), but it is semantically more correct to define the disabled value in terms of the type used by the `GAction` (`uint32`).

The alternative is to declare that the disabled value is the all-bits-set value of whatever width uint is in use at the time, but doing that would at least require you to add a comment documenting that.

andrunko: Thanks, that is fine, I thought it would make it clear to have everything as `guint` and only the GActions as `guint32` but I agree with you this is not needed as part of the fix.

Updated patch to only set selected_age to -1 on constructor.

PR #186, opened 2019-07-01
parental controls: Set max char limit to browser section help label

andrunko:
https://phabricator.endlessm.com/T27064


andrunko:
Used the same value used by the [add user dialog](https://github.com/endlessm/gnome-control-center/blob/master/panels/user-accounts/cc-add-user-dialog.ui#L534) from the same panel.

PR #190, opened 2019-07-30
applications: hide "Get ____" launchers

wjt:
These are not "real" applications, so we should hide them from the
Applications panel (just as we do for web apps).

https://phabricator.endlessm.com/T27105

### DIFF
--- a/panels/applications/cc-applications-panel.c
+++ b/panels/applications/cc-applications-panel.c
@@ -1405,6 +1405,21 @@ update_panel (CcApplicationsPanel *self,
   }
 }
 
+/* Checks whether the given @info is a pseudo-application from
+ * eos-application-tools (which opens the app center if the app is not
+ * installed, or forward to the app if it is).
+ */
+static gboolean
+has_replaced_by (GAppInfo *info)
+{
+  g_autofree gchar *replaced_by = NULL;
+
+  if (G_IS_DESKTOP_APP_INFO (info))
+    replaced_by = g_desktop_app_info_get_string (G_DESKTOP_APP_INFO (info), "X-Endless-Replaced-By");
+
+  return replaced_by != NULL;
+}
+
 static void
 populate_applications (CcApplicationsPanel *self)
 {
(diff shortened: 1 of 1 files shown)


panels/applications/cc-applications-panel.c
```diff
@@ -1425,7 +1438,8 @@ populate_applications (CcApplicationsPanel *self)
       id = get_app_id (info);
       if (!g_app_info_should_show (info) ||
           !mct_app_filter_is_appinfo_allowed (self->app_filter, info) ||
-          g_str_has_prefix (id, EOS_LINK_PREFIX))
+          g_str_has_prefix (id, EOS_LINK_PREFIX) ||
+          get_replaced_by (info) != NULL)
```
GeorgesStavracas: This is leaking the returned string.

panels/applications/cc-applications-panel.c
```diff
@@ -1405,6 +1405,19 @@ update_panel (CcApplicationsPanel *self,
...
+ * app is not installed, or forward to the app if it is. Fetch the key which
+ * indicates which app replaces @info, if any, or %NULL otherwise.
+ */
+static char *
+get_replaced_by (GAppInfo *info)
+{
+  if (G_IS_DESKTOP_APP_INFO (info))
+    return g_desktop_app_info_get_string (G_DESKTOP_APP_INFO (info), "X-Endless-Replaced-By");
+
+  return NULL;
```
GeorgesStavracas: Do you plan to use this function somewhere else? If not, I think returning a boolean and inverting the checked condition makes more sense. Something like:

```c
static gboolean
get_replaced_by (GAppInfo *info)
{
  g_autofree gchar *replaced_by = NULL;

  if (!G_IS_DESKTOP_APP_INFO (info))
    return FALSE;

  replaced_by = g_desktop_app_info_get_string (G_DESKTOP_APP_INFO (info), "X-Endless-Replaced-By");
  return replaced_by != NULL;
}
```

panels/applications/cc-applications-panel.c
wjt: Good catch.

panels/applications/cc-applications-panel.c
wjt: Yep, that's a good idea. I used a different name and slightly different (but equivalent) control flow.

PR #191, opened 2019-08-23
Translate more things

wjt:
I noticed while looking into translations for 3.6.3 that a couple of parental controls-related files were missing from POTFILES.in. Re-add these, regenerate the .pot file, and round-trip to Transifex to pick up the existing translations.

https://phabricator.endlessm.com/T27087


wjt:
Reviewed IRL by @pwithnall

PR #196, opened 2019-09-24
Fix desktop and lock screen previews to look like EndlessOS

ptomato:
https://phabricator.endlessm.com/T27897


wjt:
This looks plausible to me but I'm not super-familiar with this code.

PR #198, opened 2019-10-03
display: mark 3 missed translatable strings

wjt:
While checking that our Transifex integration is set up for this
project, I noticed that some human-facing strings were not marked for
translation - because we changed one downstream and it didn't change the
`.pot` file.

I submitted a fix upsteam at https://gitlab.gnome.org/GNOME/gnome-control-center/merge_requests/582 and backported it here. Because it conflicts with our downstream change to the string in question, I reverted that change, backported the upstream change, then reapplied the downstream string change. My hope is that this will make the next rebase easier :-)

https://phabricator.endlessm.com/T27764

### DIFF
--- a/panels/display/cc-display-panel.ui
+++ b/panels/display/cc-display-panel.ui
@@ -219,8 +219,8 @@
                                           <object class="HdyComboRow" id="primary_display_row">
                                             <property name="visible">True</property>
                                             <property name="can_focus">True</property>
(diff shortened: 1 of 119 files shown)


ptomato:
Seems fine to me, I think this is fine to merge even if the upstream one hasn't gone in yet. I doubt there will be any objections to it, it already has a thumbs up from a translation team member.

PR #204, opened 2020-01-13
T28600 unresponsive polkit dialog (temp fix)

mwleeds:
In the Users panel, a user can view their own parental controls
settings, such as what apps they can use, but non-admin users cannot
view other user's parental controls info, unless the panel is unlocked
by an admin using the "Unlock" button in the top right. This is because
in /usr/share/polkit-1/rules.d/com.endlessm.ParentalControls.rules we
only allow admins to view other user's parental controls. In case a
non-admin user is viewing another user's account, we currently try to
get the parental controls info with mct_manager_get_app_filter() even
though it won't be successful, because we can just not show that
information on failure. However due to a bug in accountsservice, an
unresponsive polkit dialog appears in this case despite libmalcontent
not using G_DBUS_CALL_FLAGS_ALLOW_INTERACTIVE_AUTHORIZATION. So as a
short term fix on the stable branch only, don't call
mct_manager_get_app_filter() if it would fail. The long term fix is to
patch GLib and accountsservice; see the ticket.

https://phabricator.endlessm.com/T28600

### DIFF
--- a/panels/user-accounts/cc-app-permissions.c
+++ b/panels/user-accounts/cc-app-permissions.c
@@ -24,6 +24,8 @@
 #include <gio/gdesktopappinfo.h>
 #include <glib/gi18n.h>
 #include <strings.h>
+#include <pwd.h>
+#include <grp.h>
 
 #include "gs-content-rating.h"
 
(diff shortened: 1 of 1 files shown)


panels/user-accounts/cc-app-permissions.c
```diff
@@ -309,13 +311,78 @@ flush_update_blacklisted_apps (CcAppPermissions *self)
     }
 }
 
+static gboolean
+current_user_in_sudo_group ()
```
pwithnall:
```suggestion
current_user_in_sudo_group (void)
```

PR #206, opened 2020-02-19
Add launcher to launch parental controls app

uajain:
Revert the existing parental control UI and add a launcher to launch the parental controls app.
___

Peek 2020-02-19 21-19

___

https://phabricator.endlessm.com/T28738


panels/user-accounts/cc-user-panel.ui
```diff
@@ -438,6 +438,48 @@
...
+                            <property name="visible">True</property>
+                            <property name="halign">GTK_ALIGN_START</property>
+                          </object>
+                        </child>
+                      </object>
+                      <packing>
+                        <property name="left_attach">1</property>
+                        <property name="top_attach">7</property>
+                        <property name="width">1</property>
+                        <property name="height">1</property>
```
pwithnall: Maybe add `<property name="x-options"></property>` to set the button so it doesn’t expand or fill? It looks a bit weird that wide.

uajain: I agree with this but probably also need to take a look at the other similar buttons as well for e.g. `Last login` and `Password:`

Screenshot from 2020-03-04 16-38-45

I will not address this here but probably will try to include the change in the 3.36-rebase branch itself.
Thanks for pointing it out.

uajain: FYI,

the new UI for user-accounts panel looks something like this:
Screenshot from 2020-03-04 16-51-15


So will adapt this particular patch to it.

pwithnall: There’s a task for adapting this patch to 3.36 [here](https://phabricator.endlessm.com/T29440), and it includes links to the upstream designs for exactly this button :)

PR #210, opened 2020-03-19
GNOME Settings should always use evince to view Endless terms of use

uajain:
Libreoffice Draw started handling `application/pdf` mimetypes
hence, the "Endless Terms of use" pdf-uri started spawning
libreoffice instead of evince. This is happening because of [1]:
"Flatpak .desktop exports take precedence over system ones due to
the order of XDG_DATA_DIRS"
    
Workaround this situation by always spawning evince for viewing
Endless Terms of use, since it's a system (non-uninstall-able)
app.

[1] : https://cgit.freedesktop.org/libreoffice/core/tree/solenv/bin/assemble-flatpak.sh#n29 

(No associated task)

### DIFF
--- a/panels/common/cc-util.c
+++ b/panels/common/cc-util.c
@@ -306,7 +306,7 @@ find_terms_document_for_languages (const gchar *product_name,
 }
 
 gboolean
-cc_util_show_endless_terms_of_use (GtkWidget *widget)
+cc_util_show_endless_terms_of_use (void)
 {
   g_autofree gchar *path = NULL;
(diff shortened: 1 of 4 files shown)


uajain:
Let's hold this for a while. 

Another approach brought to light on  https://phabricator.endlessm.com/T29587

uajain:
As discussed on slack, this is not a good approach(because requires hardcoding) as we have found a very promising solution in https://phabricator.endlessm.com/T29587

PR #215, opened 2020-03-25
Fix UI glitches

andrunko:
Also add a separate view for when metrics is disabled, similar to other panels.

Note that we don't have a metrics icon and neither a metrics-disabled icon so I went with "action-unavailable-symbolic" which looks quite Ok as a temp solution.

This PR includes the fixup from @uajain + a few other fixes, I pushed everything as a single fixup! commit on top of Umang changes as they are all fixups anyway, but let me know if you prefer to split the changes in separate commits.

https://phabricator.endlessm.com/T29571

### DIFF
--- a/panels/metrics/cc-metrics-panel.ui
+++ b/panels/metrics/cc-metrics-panel.ui
@@ -5,100 +5,148 @@
   <template class="CcMetricsPanel" parent="CcPanel">
     <property name="visible">True</property>
     <child>
-      <object class="GtkScrolledWindow">
-        <property name="visible">1</property>
-        <property name="hscrollbar-policy">never</property>
+      <object class="GtkStack" id="stack">
+        <property name="visible">true</property>
         <child>
-          <object class="HdyColumn">
-            <property name="visible">True</property>
-            <property name="maximum_width">600</property>
-            <property name="linear_growth_width">400</property>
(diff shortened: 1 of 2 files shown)


uajain:
Can't review my own commit (hope Andre has done that) but this PR is basically LGTM from my side.

panels/metrics/cc-metrics-panel.ui
```diff
@@ -5,100 +5,148 @@
...
+                <property name="label" translatable="yes">Metrics services turned off</property>
+                <attributes>
+                  <attribute name="scale" value="1.44" />
+                </attributes>
+              </object>
+            </child>
+            <child>
+              <object class="GtkLabel">
+                <property name="visible">true</property>
+                <property name="label" translatable="yes">No applications can obtain metrics information.</property>
```
wjt: This is not what turning metrics off means.

wjt:
```suggestion
                <property name="label" translatable="yes">No anonymous usage data will be sent to Endless.</property>
```

perhaps.

uajain: Addressed

andrunko:
Force pushed to properly pass "error" to g_dbus_proxy_call_finish.

PR #216, opened 2020-03-25
eos: Add a transitional message about parental controls for admins

pwithnall:
EOS 3.7 and older supported setting parental controls on administrator
accounts, as that was how Hack was set up (the child user account was an
administrator). EOS 3.8 supports creating a separate administrator
account and leaving the child (main) user as a non-administrator.
Support for setting parental controls on an administrator account has
been dropped.

This commit adds a transitional message to direct users towards the
relevant FAQ entry about the change, which will give them information
about how to reconfigure their users so they can continue to apply
parental controls to their child.

This should not be upstreamed. It can be dropped from EOS 3.9 onwards.

https://phabricator.endlessm.com/T29385

### DIFF
--- a/panels/user-accounts/cc-user-panel.ui
+++ b/panels/user-accounts/cc-user-panel.ui
@@ -328,6 +328,18 @@
                                           </packing>
                                         </child>
                                         <child>
+                                          <object class="GtkImage" id="parental_control_information">
+                                            <property name="visible">False</property>
+                                            <property name="icon-name">dialog-information-symbolic</property>
+                                            <style>
+                                              <class name="dim-label"/>
+                                            </style>
+                                          </object>
+                                          <packing>
+                                            <property name="pack_type">end</property>
+                                          </packing>
+                                        </child>
+                                        <child>
                                           <object class="GtkLabel" id="parental_controls_button_label">
                                             <property name="visible">True</property>
                                             <style>
@@ -654,5 +666,18 @@
       <widget name="autologin_row"/>
       <widget name="last_login_row"/>
     </widgets>
+  </object>
+  <object class="GtkPopover" id="parental_controls_unavailable_popover">
+    <property name="can_focus">False</property>
+    <property name="border_width">12</property>
+    <child>
+      <object class="GtkLabel">
+        <property name="visible">True</property>
+        <property name="can-focus">False</property>
+        <property name="wrap">True</property>
+        <property name="label" translatable="yes">Parental controls can now only be applied to non-administrator accounts. See the &lt;a href="https://community.endlessos.com/t/how-have-parental-controls-changed-in-endless-os-3-8/12313"&gt;FAQ entry&lt;/a&gt;.</property>
+        <property name="use-markup">True</property>
+      </object>
+    </child>
   </object>
 </interface>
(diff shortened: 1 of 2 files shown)


pwithnall:
This can’t be merged until we’ve got an FAQ entry in place for it. I’ll write one on the Phabricator task tomorrow.

Here’s how it looks. The popover appears when you click on the row while it says ‘Unavailable’. The appearance of the row isn’t changed for non-admin users.

pwithnall:
> Changes lgtm, I am just not 100% sure if we should go with an always visible message or a popover, but I am fine with the popover as well.

I thought an always-visible message would have been harder to integrate nicely into the rest of the UI. The popover appears if you click anywhere on the ‘Parental Controls’ row, so I think people will notice it.

I’ve just updated the commit to only show the popover if the admin account actually has some parental controls enabled. If they don’t, the row just says ‘Unavailable’ without the ‘i’ icon or popover.

panels/user-accounts/cc-user-panel.ui
```diff
@@ -655,4 +667,17 @@
...
   </object>
+  <object class="GtkPopover" id="parental_controls_unavailable_popover">
+    <property name="can_focus">False</property>
+    <property name="border_width">12</property>
+    <child>
+      <object class="GtkLabel">
+        <property name="visible">True</property>
+        <property name="can-focus">False</property>
+        <property name="wrap">True</property>
+        <property name="label" translatable="yes">Parental controls can now only be applied to non-administrator accounts. See the &lt;a href="https://support.endlessm.com/?TODO"&gt;FAQ entry&lt;/a&gt;.</property>
```
uajain: This can cause bugs like https://phabricator.endlessm.com/T29270

Mind adapting the approach? : https://github.com/endlessm/gnome-control-center/blob/master/panels/info-overview/cc-info-overview-panel.ui#L169-L181

pwithnall: Because translators could mis-translate the markup?

uajain: Yes, I like to follow "no tags in the translatable string" policy; as long as we have got a alternative.

pwithnall: I think that would make for a clunkier UI and would prefer to instead ensure that translators don’t break the markup (we already have to trust them to get format placeholders correct, and they can break an application just as easily if mis-translated).

uajain: Ok, just a personal preference thing. :)

pwithnall:
Rebased and updated to link to the final FAQ entry. Ready for final review.

panels/user-accounts/cc-user-panel.ui
```diff
@@ -655,4 +667,17 @@
...
   </object>
+  <object class="GtkPopover" id="parental_controls_unavailable_popover">
+    <property name="can_focus">False</property>
+    <property name="border_width">12</property>
+    <child>
+      <object class="GtkLabel">
+        <property name="visible">True</property>
+        <property name="can-focus">False</property>
+        <property name="wrap">True</property>
+        <property name="label" translatable="yes">Parental controls can now only be applied to non-administrator accounts. See the &lt;a href="https://community.endlessos.com/t/how-have-parental-controls-changed-in-endless-os-3-8/12313"&gt;FAQ entry&lt;/a&gt;.</property>
```
wjt, line 678: I worry slightly about translations of the URL but I think in practice we will have it in English and manually adjust the translations in other languages as and when someone translates it on the forum.

PR #217, opened 2020-04-13
user-accounts: Use custom setting to override faces

danigm:
This patch provides an easy way to override default faces using
org.gnome.desktop.interface.facesdirs settings configuration that can be
used in gnome-initial-setup too so downstream can override default
avatar faces without the need of a patch.

This is a Downstream patch of https://gitlab.gnome.org/GNOME/gnome-control-center/-/merge_requests/553

https://phabricator.endlessm.com/T29718

### DIFF
--- a/panels/user-accounts/cc-avatar-chooser.c
+++ b/panels/user-accounts/cc-avatar-chooser.c
@@ -411,17 +411,103 @@ cheese_camera_device_monitor_new_cb (GObject *source,
 }
 #endif /* HAVE_CHEESE */
 
+static GStrv
+get_settings_facesdirs (void)
+{
+        g_autoptr(GPtrArray) facesdirs = g_ptr_array_new ();
+        g_autoptr(GSettings) settings = g_settings_new ("org.gnome.desktop.interface");
+        g_auto(GStrv) settings_dirs = g_settings_get_strv (settings, "avatar-directories");
+
+        if (settings_dirs != NULL) {
+                int i;
+                for (i = 0; settings_dirs[i] != NULL; i++) {
+                        char *path = settings_dirs[i];
+                        if (path != NULL && g_strcmp0 (path, "") != 0)
+                                g_ptr_array_add (facesdirs, g_strdup (path));
+                }
+        }
+        g_ptr_array_add (facesdirs, NULL);
+
+        return (GStrv) g_steal_pointer (&facesdirs->pdata);
+}
+
+static GStrv
+get_system_facesdirs (void)
+{
+        g_autoptr(GPtrArray) facesdirs = NULL;
+        const char * const * data_dirs;
+        int i;
+
+        facesdirs = g_ptr_array_new ();
+
+        data_dirs = g_get_system_data_dirs ();
+        for (i = 0; data_dirs[i] != NULL; i++) {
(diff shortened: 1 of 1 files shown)


panels/user-accounts/cc-avatar-chooser.c
```diff
@@ -411,77 +411,133 @@ cheese_camera_device_monitor_new_cb (GObject *source,
...
+                                char *path = settings_dirs[i];
+                                if (path != NULL && g_strcmp0 (path, "") != 0)
+                                        g_ptr_array_add (facesdirs, g_strdup (path));
+                        }
+                }
+        }
 
-        g_signal_connect (self->flowbox, "child-activated",
-                          G_CALLBACK (face_widget_activated), self);
+        return (GStrv) g_steal_pointer (&facesdirs->pdata);
```
wjt, line 446: https://gitlab.gnome.org/GNOME/gnome-control-center/-/merge_requests/553/diffs#note_770390

panels/user-accounts/cc-avatar-chooser.c
```diff
@@ -411,77 +411,133 @@ cheese_camera_device_monitor_new_cb (GObject *source,
...
+                        if (target != NULL && g_str_has_prefix (target , "legacy/"))
                                 continue;
-                        }
 
-                        file = g_file_get_child (dir, g_file_info_get_name (info));
-                        g_list_store_append (self->faces, file);
-
-                        g_object_unref (info);
+                        face_file = g_file_get_child (dir, g_file_info_get_name (info));
+                        g_list_store_append (faces, face_file);
```
andrunko, line 506: Don't we need to use g_steal_pointer here for face_file? Or not use g_autoptr for it.

panels/user-accounts/cc-avatar-chooser.c
```diff
@@ -411,77 +411,133 @@ cheese_camera_device_monitor_new_cb (GObject *source,
...
+
+        g_signal_connect (self->flowbox, "child-activated",
+                          G_CALLBACK (face_widget_activated), self);
+
+        facesdirs = get_settings_facesdirs ();
+        added_faces = add_faces_from_dirs (self->faces, facesdirs, TRUE);
+
+        if (!added_faces) {
+                facesdirs = get_system_facesdirs ();
+                add_faces_from_dirs (self->faces, facesdirs, FALSE);
```
andrunko, line 539: Any reason to not always include faces from system dirs instead of only when `added_faces` is false? Similarly, why not pass TRUE to `add_all` here? If that is the case we could remove the `add_all` param altogether from `add_faces_from_dirs`.

panels/user-accounts/cc-avatar-chooser.c
andrunko: Clarified upstream, g_list_store_append takes a ref of the arg.

panels/user-accounts/cc-avatar-chooser.c
andrunko: Also clarified upstream.

panels/user-accounts/cc-avatar-chooser.c
danigm: I've updated the PR adding a NULL pointer to the array.

PR #220, opened 2020-05-12
fixup! user-accounts: Use custom setting to override faces

andrunko:
https://phabricator.endlessm.com/T30095


GeorgesStavracas:
Looks good.

PR #227, opened 2020-09-29
privacy: Mention that system apps can always use the camera/microphone

pwithnall:
These toggles only apply to the portal system, so only restrict access
for sandboxed apps. System apps (installed without sandboxing into
`/usr/bin`) can always access these resources on typical distributions.

In future, perhaps integration with a MAC (like SELinux) could restrict
access to the camera/microphone/GPS for system applications — but we
invented sandboxing to avoid having to do that, so it seems unlikely
that’s ever going to be implemented. If it is, this wording can change
again.

Upstream: https://gitlab.gnome.org/GNOME/gnome-control-center/-/merge_requests/851
Signed-off-by: Philip Withnall <pwithnall@endlessos.org>

Fixes: #741

https://phabricator.endlessm.com/T27784

### DIFF
--- a/panels/camera/cc-camera-panel.ui
+++ b/panels/camera/cc-camera-panel.ui
@@ -72,7 +72,7 @@
                       <object class="GtkLabel">
                         <property name="visible">true</property>
                         <property name="margin-bottom">12</property>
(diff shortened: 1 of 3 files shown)


andrunko (approved):
This lgtm although I'm not sure regular users will understand the difference between system apps and flatpaks... but I have no better idea - maybe @wjt has some suggestion?

wjt:
I do not

PR #229, opened 2020-10-22
updates: Send a metric when automatic updates settings change

pwithnall:
This will give us some measure of how many people are using automatic
updates, and how many of those are on metered connections.

Signed-off-by: Philip Withnall <pwithnall@endlessos.org>

https://phabricator.endlessm.com/T29248


GeorgesStavracas (approved):
Only a pair of nitpicks, please merge after either addressing or rejecting the comments.

PR #244, opened 2022-01-19
41.2

GeorgesStavracas:
### Cherry-picked

b45095df7 user-accounts: allow empty/weak passwords
ccf75f55c panels: lock: Disable lock screen when autologin is enabled
cb4364df4 user-accounts: Fall back to "user" as the default login name
e9d41b471 user-accounts: Allow account creation when name contains no ASCII
a6c9969d4 user-accounts: Allow to create a user without a password
47150fb52 user-accounts: Allow to remove the password for a user
05bfae6cc user-accounts: Show the correct password mode when changing it
7f2e4c201 user-accounts: Allow to change the password mode for oneself
45b36921e Populate language chooser with FBE vendor config file
48270a13b info-overview: add link for license attribution document
65f2434b2 user panel, password: add fallback kbd layout if necessary
89631fb89 updates: Introduce Automatic Updates panel
9265873b3 Don't declare StartupNotify=true in panel-specific desktop files
4e62bd247 panels: Add X-GNOME-UsesNotifications to several desktop files
06e797c4b user-accounts: Limit to 80 characters
6bbbdce63 applications: Use "Open in App Center" instead of "Open in Software"
c3a52f9d9 applications: hide “Show in App Center” for App Center
bcf9fd670 privacy: Add metrics panel controlling metrics service integration
  * Squash with 7fad7216f metrics: Remove metrics identifier

2ddc19518 Debian's adduser doesn't allow uppercase letters by default.
c1d7147bd privacy: Mention that system apps can always use the camera/microphone
f17c37434 eos: Add metric for opening a panel
e241d5ce5 updates: Factor out a variable
e67553c2e updates: Send a metric when automatic updates settings change
8f7192de7 Add translation support for several new languages
94e995d06 Add Transifex translation infrastructure

### Drop

fa10c2277 region: update the locale format along with the language when it changes
 * Upstreamed

6cd113b33 bluetooth: set switch off when BT airplane mode is enabled
 * Similar upstream change

7c0a5bf1d eos: Move the shell bar to the bottom of the background preview
 * EOS5 panel is at the top

4f1ff7f83 background: also monitor the wallpapers directory
 * BgPicturesSource was removed in 40.0

e595866b3 user-accounts: Fix a critical warning if loading a cursor fails
 * Fixed upstream

-----

https://phabricator.endlessm.com/T32957

### DIFF
--- a/.tx/config
+++ b/.tx/config
@@ -1,0 +1,9 @@
+[main]
+host = https://www.transifex.com
+
+[gnome-control-center.app]
+source_file = po/gnome-control-center-2.0.pot
+source_lang = en
+file_filter = po/<lang>.po
+type = PO
+


pwithnall (commented):
* `region: update the locale format along with the language when it changes` seems to almost be a no-op.
 * `panels: Add X-GNOME-UsesNotifications to several desktop files` — is this still relevant? If so, the commit message needs to be updated since it only changes one panel now instead of five.
 * `applications: hide “Show in App Center” for App Center` — could this be upstreamed by changing it to show ‘Open’ instead of ‘Open in Software’ for gnome-software? Then it’s decoupled from EOS’ behaviour of not showing gnome-software inside itself.

GeorgesStavracas:
> region: update the locale format along with the language when it changes seems to almost be a no-op.

You're right, this patch has already been upstreamed.

> panels: Add X-GNOME-UsesNotifications to several desktop files — is this still relevant? If so, the commit message needs to be updated since it only changes one panel now instead of five.

I wonder why this patch differed from https://gitlab.gnome.org/GNOME/gnome-control-center/-/merge_requests/265. I'll push this upstream.

> applications: hide “Show in App Center” for App Center — could this be upstreamed by changing it to show ‘Open’ instead of ‘Open in Software’ for gnome-software? Then it’s decoupled from EOS’ behaviour of not showing gnome-software inside itself.

The Applications panel was redesigned and the new design doesn't mention "Software": https://gitlab.gnome.org/Teams/Design/settings-mockups/-/blob/master/applications/applications-wires.png

This is probably the last rebase we'll see this patch. Let's treat it with love and respect, since it's so close to retiring :older_adult: :heart:

pwithnall:
> The Applications panel was redesigned and the new design doesn't mention "Software": https://gitlab.gnome.org/Teams/Design/settings-mockups/-/blob/master/applications/applications-wires.png

The new upstream implementation explicitly hides gnome-software from the ‘Applications’ list?

GeorgesStavracas:
> The new upstream implementation explicitly hides gnome-software from the ‘Applications’ list?

No, Software is shown. I was talking about the "Open in Software" button, that was renamed to "View Details"

pwithnall:
> > The new upstream implementation explicitly hides gnome-software from the ‘Applications’ list?
> 
> No, Software is shown. I was talking about the "Open in Software" button, that was renamed to "View Details"

The g-c-c patch is to hide the ‘Software’ entry from the applications list, though, because EOS’ gnome-software [hides itself in its own application listings](https://github.com/endlessm/gnome-software/blob/a73f24a50dbffadfe36f038b66d4f8dc4aa573eb/plugins/eos-blocklist/gs-plugin-eos-blocklist.c#L844). So on EOS the ‘View Details’ or ‘Open in Software’ button will open gnome-software only to display an error message along the lines of “org.gnome.Software not found as an app”.

GeorgesStavracas:
> The g-c-c patch is to hide the ‘Software’ entry from the applications list, though, because EOS’ gnome-software hides itself in its own application listings. So on EOS the ‘View Details’ or ‘Open in Software’ button will open gnome-software only to display an error message along the lines of “org.gnome.Software not found as an app”.

Right, sorry, I got confused with what the patches. The patch that will be retired is ` 6bbbdce applications: Use "Open in App Center" instead of "Open in Software"`. I don't think we'll be able to drop `applications: hide “Show in App Center” for App Center` even with the redesign. Do notice, thought, that the redesigned Applications panel contains a button to launch the app, which addresses your original concern.

pwithnall:
Thanks for checking. LGTM to merge now

PR #245, opened 2022-04-29
Revert "applications: hide “Show in App Center” for App Center"

pwithnall:
This reverts commit 13c14a3c7a5983f8deb2ee895fec22e337c33ca8.

As of T33045, the change in downstream gnome-software to make it hide
itself has been dropped. This means that the “Show in App Center” button
should now work for gnome-software in gnome-control-center.

This commit and the one it reverts can be dropped next time we rebase.

https://phabricator.endlessm.com/T33045


wjt (approved):
Lgtm, feel free to merge after presumably picking an upstream patch to fix the meson usage

PR #248, opened 2023-12-04
Rebase on 43.6

dbnicholson:
The rebased series can be seen [here](https://github.com/endlessm/gnome-control-center/compare/Version_43.6...T35056-rebase-43.6). Debian branch PR in #249. Here's a list of the previous commits:

- [ ] 5669987a7 Post-release version bump
- [ ] 492fac154 Update Portuguese translation
- [ ] 293e191e3 display-config: Do not invert order when constructing modes
- [ ] ad29e2677 Update Russian translation
- [ ] 0e2562c2c applications: Set the Icon key in desktop file to a icon that exists
- [ ] 8d318a722 applications: Fix typo in key name in the desktop file
- [ ] 4f64deb5f meson: drop unused argument for i18n.merge_file()

Drop, upstream backports on the gnome-41 branch.

- [x] 77d5c22c6 user-accounts: allow empty/weak passwords
- [x] 9fe54567e panels: lock: Disable lock screen when autologin is enabled
- [x] d540fbf07 user-accounts: Fall back to "user" as the default login name
- [x] bd13c0eae user-accounts: Allow account creation when name contains no ASCII
- [x] 4f67d0fc0 user-accounts: Allow to create a user without a password
- [x] 875216a12 user-accounts: Allow to remove the password for a user
- [x] dc9bbe170 user-accounts: Show the correct password mode when changing it
- [x] a546fef10 user-accounts: Allow to change the password mode for oneself

Keep

- [x] b6c86d8be Populate language chooser with FBE vendor config file

Keep? Kinda seems like the type of thing we don't want to bother with
anymore, but I don't know if we have custom images using this setup.

- [x] b5d855992 info-overview: add link for license attribution document

Keep

- [ ] 93bf79268 user panel, password: add fallback kbd layout if necessary

Drop https://phabricator.endlessm.com/T31812

- [ ] 60345b3f0 updates: Introduce Automatic Updates panel

Deferred for gtk4/libadwaita port.

- [ ] 401714504 Don't declare StartupNotify=true in panel-specific desktop files

Drop. This was fixed in gtk
(https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/156) and backported
to all stable branches.

- [ ] d68ff3674 panels: Add X-GNOME-UsesNotifications to several desktop files

Drop. The notifications emitted for the universal access panel came from
the gnome-settings-daemon keyboard accessibility plugin, but that's been
removed.

- [x] ea1cc373f user-accounts: Limit to 80 characters

Keep

[...]

PR #249, opened 2023-12-04
Rebase on Debian 43.6-2~deb12u1

dbnicholson:
The rebased series can be seen [here](https://github.com/endlessm/gnome-control-center/compare/Version_43.6_debian...T35056-rebase-43.6-debian). Master branch PR in #248. Here's a list of the previous commits:

- [ ] 6ac86b6e1 Remove .pkla file

Drop. We certainly don't need the PKLA file, but leaving it in `/var` is
harmless and gets either deleted from the ostree or masked at runtime,
anyways.

- [x] 9e0893c46 Add build dependency on eos-metrics-0-dev
- [x] 8f00ac287 Add libmogwai-tariff-0 dependency
- [x] cb56ae545 control: Drop desktop-base dependency

Keep

The following commit is added:

- b4ef6a6677 Disable snap integration on Endless

The build currently fails without a patch to change the dependency to
snapd-glib-2, but we're not shipping snap support, anyways.

Status of Debian's patches:

- [x] [keyboard-Allow-disabling-alternate-characters-key.patch](debian/patches/keyboard-Allow-disabling-alternate-characters-key.patch)

Applied

- [ ] [build-Look-for-snapd-glib-2.patch](debian/patches/build-Look-for-snapd-glib-2.patch)

Dropped. Snap integration disabled as noted above.

- [ ] [debian/Expose-touchpad-settings-if-synaptics-is-in-use.patch](debian/patches/debian/Expose-touchpad-settings-if-synaptics-is-in-use.patch)

Dropped. We don't ship the synaptics Xorg driver.

- [x] [debian/Debian-s-adduser-doesn-t-allow-uppercase-letters-by-defau.patch](debian/patches/debian/Debian-s-adduser-doesn-t-allow-uppercase-letters-by-defau.patch)

Applied

- [ ] [debian/Ignore-result-of-test-network-panel.patch](debian/patches/debian/Ignore-result-of-test-network-panel.patch)

Dropped. This only seems to affect very slow architectures.

- [ ] [debian/firmware-security-disable.patch](debian/patches/debian/firmware-security-disable.patch)

Dropped. Unnecessary divergence from upstream. We may want this, though,
as we don't have fwupd installed and all the data will be missing.

https://phabricator.endlessm.com/T35056

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -20,7 +20,11 @@ endif
 endif
 
 ifeq ($(DEB_HOST_ARCH_OS),linux)
+ifeq ($(shell dpkg-vendor --is Endless && echo yes),yes)
+SNAP := -Dsnap=false
+else
 SNAP := -Dsnap=true
+endif
 else
 SNAP := -Dsnap=false
 endif


dylanmccall (approved):
🥳

PR #250, opened 2023-12-11
search-provider: Show Parental Controls (if installed)

wjt:
This is a bit of a hack but it works fine.

Screenshot from 2023-12-11 11-16-46

https://phabricator.endlessm.com/T35123


wjt:
This works but does not address the issue where the Parental Controls app in App Centre can't be launched. So for the time being I will revert the upstream change and unhide the launcher. [Upstream issue here](https://gitlab.freedesktop.org/pwithnall/malcontent/-/issues/48).

PR #251, opened 2023-12-13
fixup! Populate language chooser with FBE vendor config file

dbnicholson:
gnome-initial-setup has been updated to use the upstream vendor config files at `/etc/gnome-initial-setup/vendor.conf` and `/usr/share/gnome-initial-setup/vendor.conf`. Our downstream `/var/lib/eos-image-defaults/branding/gnome-initial-setup.conf` config file is no longer being used. Adjust the language list population accordingly.

This can be squashed with 39117fc374 on the next rebase.

https://phabricator.endlessm.com/T35040

I think this could probably be sent upstream since it's now using g-i-s upstream paths.

### DIFF
--- a/panels/common/cc-common-language.c
+++ b/panels/common/cc-common-language.c
@@ -218,7 +218,8 @@ insert_language (GHashTable *ht,
         }
 }
 
-#define FBE_VENDOR_CONF_FILE "/var/lib/eos-image-defaults/branding/gnome-initial-setup.conf"
+#define FBE_LOCAL_CONF_FILE "/etc/gnome-initial-setup/vendor.conf"
+#define FBE_VENDOR_CONF_FILE "/usr/share/gnome-initial-setup/vendor.conf"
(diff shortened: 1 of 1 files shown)


wjt (approved):
I had no memory of this code being here.

I think you are right that this could go upstream; but actually this language list loading code is not upstream in initial setup either. I have the vague sense that gsettings would have been better than this ad-hoc text file. Oh well.

dbnicholson:
> I think you are right that this could go upstream; but actually this language list loading code is not upstream in initial setup either. I have the vague sense that gsettings would have been better than this ad-hoc text file. Oh well.

Definitely feels like a gsetting to me.

wjt:
The problem is… the whole file feels like a series of GSettings to me.

dbnicholson:
Indeed. Out of curiosity, why was the `vendor.conf` file preferred to GSettings? Too annoying to handle with the dedicated system user (a la gdm)?

PR #254, opened 2024-03-25
fixup! privacy: Add metrics panel controlling metrics service integration

dylanmccall:
Remove redundant calls to `gtk_switch_set_state` with the metrics toggle switch. Instead, bind switch `state` directly to the value of the D-Bus object property, and set the `active` property to follow when it changes. With this change, the switch's `state` property refers to the `Enabled` property of the metrics daemon, while the `active` property refers to the user's intent. Since Gtk 4.9.3, it is necessary to control the state and active properties independently.

https://phabricator.endlessm.com/T35277

### DIFF
--- a/panels/metrics/cc-metrics-panel.c
+++ b/panels/metrics/cc-metrics-panel.c
@@ -36,9 +36,7 @@ struct _CcMetricsPanel
   AdwStatusPage      *metrics_disabled_page;
   AdwPreferencesPage *preferences_page;
 
-  gboolean       metrics_active;
   gboolean       changing_state;
-  gboolean       pending_state;
 
   GDBusProxy    *metrics_proxy;
 };
@@ -46,9 +44,9 @@ struct _CcMetricsPanel
 CC_PANEL_REGISTER (CcMetricsPanel, cc_metrics_panel)
 
 static void
-on_metrics_panel_set_enabled (GObject      *source_object,
-                              GAsyncResult *res,
-                              gpointer      user_data)
+on_metrics_proxy_set_enabled_finish (GObject      *source_object,
+                                     GAsyncResult *res,
+                                     gpointer      user_data)
 {
   CcMetricsPanel *self = CC_METRICS_PANEL (user_data);
   g_autoptr(GVariant) results = NULL;
@@ -66,15 +64,12 @@ on_metrics_panel_set_enabled (GObject      *source_object,
     }
 
   self->changing_state = FALSE;
-  self->metrics_active = self->pending_state;
-  gtk_switch_set_state (GTK_SWITCH (self->enable_metrics_switch),
-                        self->metrics_active);
 }
 
 static gboolean
-metrics_switch_active_changed_cb (GtkSwitch *widget,
-                                  gboolean   new_state,
-                                  gpointer   user_data)
+on_metrics_switch_state_set (GtkSwitch *widget,
+                             gboolean   new_state,
+                             gpointer   user_data)
 {
   CcMetricsPanel *self = CC_METRICS_PANEL (user_data);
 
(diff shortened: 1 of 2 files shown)


panels/metrics/cc-metrics-panel.c
```diff
@@ -66,30 +64,26 @@ on_metrics_panel_set_enabled (GObject      *source_object,
...
+on_metrics_switch_state_set (GtkSwitch *widget,
+                             gboolean   new_state,
+                             gpointer   user_data)
 {
   CcMetricsPanel *self = CC_METRICS_PANEL (user_data);
 
   if (self->changing_state)
     return TRUE;
 
   self->changing_state = TRUE;
```
wjt, line 79: Consider the following:

1. Open this panel
2. Ensure metrics are enabled
3. Send `SIGSTOP` to the metrics daemon so that the SetEnabled call(s) takes a long time
4. Click the switch twice
5. Send `SIGCONT` to the metrics daemon to allow it to reply to method calls

I think the expected behaviour is that clicking twice disables then reënables metrics but by my reading the result will be that metrics are disabled. Is that the desired behaviour? Not sure!

panels/metrics/cc-metrics-panel.c
```diff
@@ -190,25 +179,29 @@ cc_metrics_panel_constructed (GObject *object)
...
     metrics_can_change = g_permission_get_allowed (permission);
 
   box = gtk_box_new (GTK_ORIENTATION_HORIZONTAL, 6);
-  gtk_widget_show (box);
+  gtk_widget_set_visible (box, true);
 
   gtk_widget_set_sensitive (GTK_WIDGET (self->enable_metrics_switch), metrics_can_change);
-  gtk_switch_set_state (GTK_SWITCH (self->enable_metrics_switch), self->metrics_active);
+
+  /* It would be clearer to use our own metrics_enabled property, but for this
```
wjt, line 196: Is there such a property?

panels/metrics/cc-metrics-panel.c
```diff
@@ -190,25 +179,29 @@ cc_metrics_panel_constructed (GObject *object)
...
+   * purpose it is fine to treat the switch's state property as canonical. */
+  g_object_bind_property (self->enable_metrics_switch, "state",
+                          self->enable_metrics_switch, "active",
+                          G_BINDING_SYNC_CREATE);
```
wjt, line 200: I don't quite understand why this is needed. Is it to handle the daemon property changing behind this page's back?

I looked at how another page in this app, the Thunderbolt one in particular, handles this. There, on initialisation they block the `state-set` handler, then set the `active` property (which causes the default `state-set` handler to run which just sets `state`).

https://gitlab.gnome.org/GNOME/gnome-control-center/-/blob/3b22ae0da6aed94c8226f03194aee64a159d2653/panels/privacy/cc-bolt-page.c#L467-471

dylanmccall: Ooh, I see how they did that. That's much nicer, indeed. I switched over to that pattern in the newest commit. Unfortunately, in my testing, the default state-set handler isn't running / isn't setting the "active" property, which is strange and confusing but I unfortunately can't dig into it very well because of a sad combination of slow internet and oddly sluggish laptop. I worked around it by calling both `gtk_switch_set_state ()` and `gtk_switch_set_active ()` in a new `cc_metrics_panel_enable_sync` function, which is lame but at least contains the weird hacky stuff to one place. If it bothers you, I can take a better look on Wednesday when I'm back in Canada with a big computer.

panels/metrics/cc-metrics-panel.c
dylanmccall: I was frowning at that as well but didn't want to mess with it too much. (I think what you describe matches the current behaviour, and other panels like the Location and Camera panels do the same odd thing). I guess we _could_ actually make use of GCancellable here, but I don't think it's a regression.

panels/metrics/cc-metrics-panel.c
dylanmccall: There isn't. I rearranged this in such a way that it's more clear what is going on, so that awkward comment is unnecessary :)

wjt (approved):

